PR #27, opened 2017-05-15
MacOS support for 3.4.5... 

sigysmund:
MacOS(sierra) development environment.
Updated qmake project configs.
Deploy & Run script.
For now only hardware supported is HackRF.
Quick HowToBuild.

---

Aware: I did not check other operating-system building capabilities (But it should not break :) ... )


f4exb:
Yes it is working fine for the other builds. Merged to master and fix and also to dev branch for the coming v3.5.0.

Thanks a lot for your contribution!

sigysmund:
While using SDRangel, I have noticed some occasions, where there are problems needed to be fixed.
(What an amusing thing to say...) 

I will fix it when I am enough annoyed and know exactly what is causing differences,
between my usage of app in linux(where there is stable).

Version 2.0.0 with El Capitan seemed more stable as 3.4.5 with Sierra :)

Thank you for many great pieces of software!

PR #164, opened 2018-04-22
Memleak fixes

jigpu:
Running sdrangel through valgrind reveals a number of memory leaks. I'm having difficulty tracking most of them down, but the two patches in this pull request fix two pretty easy ones.


jigpu:
My apologies. Didn't see any issue locally, so I'm not sure what went wrong...

f4exb:
Problem is in on_action_Exit_triggered(). I don't know when "about to exit" signal is emitted but it looks like it is too early.

PR #183, opened 2018-05-31
Fix build on current Arch Linux

jigpu:
After updating my Arch Linux system I was no longer able to compile sdrangel. It seems that the issue is due to updated versions of Qt 5 and ffmpeg. The former has removed a deprecated cmake macro that we are using and the latter has renamed a symbol that we were relying on.

The second patch should probably be rewritten to so that it is compatible with both old and new versions of ffmpeg (and add appropriate cmake requirements), but that's a decision for the maintainer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,7 +56,8 @@ set(CMAKE_AUTOMOC ON)
 find_package(Qt5Core 5.0 REQUIRED)
 find_package(Qt5Widgets 5.0 REQUIRED)
 find_package(Qt5Multimedia 5.0 REQUIRED)
-#find_package(QT5OpenGL 5.0 REQUIRED)
+find_package(Qt5MultimediaWidgets 5.0 REQUIRED)
+find_package(Qt5OpenGL 5.0 REQUIRED)
 find_package(OpenGL REQUIRED)
 find_package(PkgConfig)
 find_package(Boost REQUIRED)
@@ -276,7 +277,7 @@ if(WIN32)
     set_target_properties(sdrangel PROPERTIES LINK_FLAGS_MINSIZEREL "/SUBSYSTEM:windows /ENTRY:mainCRTStartup")
 endif(WIN32)
 
-qt5_use_modules(sdrangel Widgets Multimedia)
+target_link_libraries(sdrangel Qt5::Widgets Qt5::Multimedia)
 
 ##############################################################################
 # main server application
@@ -300,7 +301,7 @@ target_link_libraries(sdrangelsrv
     ${QT_LIBRARIES}
 )
 
-qt5_use_modules(sdrangelsrv Multimedia)
+target_link_libraries(sdrangelsrv Qt5::Multimedia)
 
 ##############################################################################
 # main benchmark application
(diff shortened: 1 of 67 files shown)


httpserver/CMakeLists.txt
```diff
@@ -50,6 +50,6 @@ target_link_libraries(httpserver
 	${QT_LIBRARIES}
 )
 
-qt5_use_modules(httpserver Core Network)
+target_link_libraries(httpserver Qt5::Core Qt5::Network)
```
f4exb: It is nicer to add these dependencies in the target_link_libraries block opened just above

CMakeLists.txt
```diff
@@ -56,7 +56,8 @@ set(CMAKE_AUTOMOC ON)
 find_package(Qt5Core 5.0 REQUIRED)
 find_package(Qt5Widgets 5.0 REQUIRED)
 find_package(Qt5Multimedia 5.0 REQUIRED)
-#find_package(QT5OpenGL 5.0 REQUIRED)
```
f4exb: I believe this has been removed to also remove dependency with graphical items at top level. This is inconsistent with the rest of required packages in this list like Widgets and OpenGL itself so it needs more careful revision. I think the intention was to open possibility to compile only a server version.

CMakeLists.txt
```diff
@@ -56,7 +56,8 @@ set(CMAKE_AUTOMOC ON)
...
+find_package(Qt5MultimediaWidgets 5.0 REQUIRED)
```
f4exb: If this is only needed by demodatv I would rather move it there. However this is not part of the target link libraries today is this new to Qt 5.11 ?

httpserver/CMakeLists.txt
jigpu: Good idea. I'll make this modification.

CMakeLists.txt
jigpu: Also sounds good. I'll move it out.

CMakeLists.txt
jigpu: I think we should have a seperate patch (set?) focused on removing top-level dependencies on graphics since that's bound to require much more careful planning than simply fixing the build like this set tries to do.

f4exb:
plugins/channelrx/chanalyzerng and plugins/channelrx/tcpsrc are now discontinued

CMakeLists.txt
FFY00: This is still required but the name of the package is wrong. Notice the uppercase `T`. This should be added as `find_package(Qt5OpenGL 5.0 REQUIRED)`.

FFY00:
I fixed the build for the last version (`4.0.0`). Here's the patch https://aur.archlinux.org/cgit/aur.git/tree/qt_use_modules.patch?h=sdrangel. The only thing missing from there is the `FF_INPUT_BUFFER_PADDING_SIZE`/`AV_INPUT_BUFFER_PADDING_SIZE` situation, which I patched directly in the [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=sdrangel).

If you want to apply the patch manually, you will need to replace `SOURCE_FOLDER` with the source directory.
If you're applying the patch in the current folder.
```
sed -i 's/SOURCE_FOLDER/./g' qt_use_modules.patch
```

f4exb:
Manually merged in v4.0.1 release

PR #263, opened 2018-12-15
Dev

GkvJeep:


sdrgui/CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ set(sdrgui_SOURCES
     gui/valuedialz.cpp
 
     dsp/scopevis.cpp
-#   dsp/scopevismulti.cpp
+    dsp/scopevismulti.cpp
```
f4exb: Why? This is of no use at this moment.

plugins/samplesource/CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ else(LIBUSB_FOUND AND LIBMIRISDR_FOUND)
 endif(LIBUSB_FOUND AND LIBMIRISDR_FOUND)
 
 find_package(SoapySDR)
-if(LIBUSB_FOUND AND SOAPYSDR_FOUND)
+if(SOAPYSDR_FOUND)
```
f4exb: Fine but if this is the only change I'd rather patch this file than do a merge

sdrgui/CMakeLists.txt
GkvJeep: Error compile msvc.

plugins/samplesource/CMakeLists.txt
GkvJeep: Many devices when using Soapysdr do not require libusb.
Libusb preinstalled in Soapyusdr.

sdrgui/CMakeLists.txt
GkvJeep: Error at the stage of linker, there is a link from other files.

plugins/samplesource/CMakeLists.txt
GkvJeep: The same is required for  plugins\samplesink\CMakeLists.txt
line 32   if(LIBUSB_FOUND AND SOAPYSDR_FOUND)

f4exb:
All changes implemented manually in v4.3.2

PR #269, opened 2019-01-02
pluto: Get LO range from device

rgetz:
When updating firmware, the devices which have AD9364s on them, get
reset to the default of a AD9363 (tuning range of 325 to 3800 MHz).
SDRAngel assumes a AD9364, and the GUI allows you to set LO settings
that the firmware doesn't support.

This ensures that does not happen, by going out to the hardware, and
querying the device to set the min/max limits on LO.

### DIFF
--- a/devices/plutosdr/deviceplutosdrbox.cpp
+++ b/devices/plutosdr/deviceplutosdrbox.cpp
@@ -713,6 +713,25 @@ bool DevicePlutoSDRBox::getTxRSSI(std::string& rssiStr, unsigned int chan)
     return get_param(DEVICE_PHY, buff, rssiStr);
 }
 
+void DevicePlutoSDRBox::getRxLORange(uint64_t& minLimit, uint64_t& maxLimit, unsigned int chan)
+{
+    // values are returned in Hz
+    qint64 stepLimit;
+    std::string rangeStr;
+
+    chan = chan % 2;
+    char buff[50];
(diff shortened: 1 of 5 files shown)


rgetz:
I wasn't sure if you wanted pull requests on dev? or master?

Feel free to nit on style - my c++ is a little rusty, and I would like to help with a few more minor patches some users have complained to me about.

Thanks in advance.

f4exb:
I'd rather parse the range string at the lowest level in devices/plutosdr/deviceplutosdrbox.cpp but I can refactor this later.

rgetz:
@f4exb : No problem - I can do that - move the string parsing to devices/plutosdr/deviceplutosdrbox.cpp:getRxLORange() and return min/max, and just pass things back in Hz, or kHz?

-Robin

f4exb:
I think the dial object uses kHz. Hz would be more universal but you will have mind the size and use 64 bit integers.

rgetz:
and - my only other question was on the preference on camel case - I saw lo some places, and LO other places.

getRxLORang

get RXloRange

?

f4exb:
LO are the initials of Local Oscillator hence the capital letters. Otherwise most variables are written in camel case. So: getRxLORange

rgetz:
Pull request updated. Let me know if there are other issues. (I'm sure I didn't do the 64-bit handling the same way you would like - I didn't want to add too many additional #includes for qint64 in the generic headers).

Feedback appreciated.

rgetz:
Thanks. 

That will help me (and others).

-Robin

PR #283, opened 2019-02-01
plutosdr: grab the RF bandwidth range from the device

rgetz:
This will use the RF bandwidth from the device, which is different
between AD9363 and AD9364.

Things are now managed like the device likes - analog low pass bandwidth
is RF (complex) bandwidth, not baseband single I or Q bandwidth.


rgetz:
I also have a patch which does the same thing for sample rate - but that forced me track down a bug in the firmware - I don't know if you want that separately, or squashed into this pull request? or wait until the firmware bug is fixed.

f4exb:
I think it is safer to do the sample rate when the firmware bug is fixed

PR #306, opened 2019-03-03
Run daemon dependencies using an Entrypoint script

adecarolis:


f4exb:
I am not applying this PR since the same changes now exist on the dev branch.

PR #326, opened 2019-04-18
cmake/clang fixes

ra1nb0w:
this PR has many commits, the most important are:
- libdsdcc/cm256cc fix include path and library folder on cmake
- libfreedv: fix c99 and c++11 inconsistency (now it build on clang)
- fcdpro/fcdproplus doesn't work on macOS because they require asound (alsa)
- generalize cpu flags discovery (same as cm256cc)
- add libopus-dev to readme (probably need some other adjustments in opensuse/fedora)

All tested on:
- Ubuntu 18.04 with all updates and Readme instructions
- macOS 10.14.4

All this changes permit to build sdrangel on a plain macOS 10.14 and XCode 10.2 with clang (probably with lower too but not tested). I also done macports packages (cm256cc, mbelib, dsdcc, codec2, freedv-gui and sdrangel too) but they are not upstream yet. Probably with the next sdrangel version. Full Soapy* are all there.

TODO for macOS platform:
- move library from ${prefix}/lib to ${prefix}/lib/sdrangel to avoid confusion
- add cpack to create automatic bundle and dmg
- fix annoying audio problem `SampleSinkFifo: overflow - dropping 65536 samples`
- check why


f4exb:
The only thing that worries me is that since already some time now FCDpro and FCPro+ do not use alsa anymore. The FCD sound device is handled by Qt audio stuff. So I am wondering what is still there from Alsa.

f4exb:
The only reference to "alsa" I can find in the project. I think this is obsolete:
```
/opt/build/sdrangel/fcdlib/fcdtraits.cpp
  20,30: const char *fcd_traits<Pro>::alsaDeviceName = "hw:CARD=V10";
  21,34: const char *fcd_traits<ProPlus>::alsaDeviceName = "hw:CARD=V20";

/opt/build/sdrangel/fcdlib/fcdtraits.h
  36,21: 	static const char *alsaDeviceName;
  54,21: 	static const char *alsaDeviceName;
  72,21: 	static const char *alsaDeviceName;
  83,62: template <fcd_type FCDType> const char *fcd_traits<FCDType>::alsaDeviceName = "";
```

f4exb:
Merged. We'll see for the Alsa thing later.

ra1nb0w:
thank you! Now, I see for alsa.
In the meanwhile, do you know how to move library from ${prefix}/lib to ${prefix}/lib/sdrangel during `make install` to avoid confusion? `install(TARGETS sdrgui DESTINATION lib/sdrangel)` is not enough because it fail to link the binaries during the install phase.

PR #327, opened 2019-04-19
re-enable fcdpro/fcdproplus on macOS and perseus fix

ra1nb0w:
I checked and now seems to works and no more error from alsa. Probably the error was library_include dependant.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
 cmake_minimum_required(VERSION 3.1.0)
-cmake_policy(SET CMP0043 OLD)
 
 if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
     message(FATAL_ERROR "SDRangel requires GCC version 4.9 or higher!")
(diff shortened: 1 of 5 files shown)


f4exb:
For Perseus you need my fork of libperseus where perseus-in.h is installed. I had a lot of issues with the libperseus from Microtelecom and it simply does not work as it is. The corresponding code cannot be simply removed. As far as I remember this was to give a chance to load the firmware when it was initially missing. The firmware is not kept in non volatile memory and has to be loaded once after the Perseus has been powered up. This still does not work 100% and I still have to close and restart SDRangel to get access to the Perseus so this is still not the end of the story.

It seems I am the only user of a Perseus with SDRangel though,

ra1nb0w:
I saw that you are using cmake for libperseus-lib and I know that is volatile but in my case seems to work well with this patch but surely you know the hardware better than me so I asked upstream what they think on enabling `perseus-in.h`. If they accept I can revert the patch.

f4exb:
I think that if you want to effectively use Perseus you have to use my fork of libperseus-lib anyway.

ra1nb0w:
Before using the main (I sent a few patches) I checked and the only difference was cmake.

f4exb:
Yes I changed my comment while you were typing

ra1nb0w:
The libperseus-lib developer said that the patch I did is correct. see https://github.com/Microtelecom/libperseus-sdr/pull/9#issuecomment-485230476
Maybe, do more tests and then you can merge. thank you

PR #328, opened 2019-04-23
udev on Linux and libusb fix

ra1nb0w:
* udev fixes 07c1413c
* usb fixes 760e4859

tested on Ubuntu 18.04 and macOS


ra1nb0w:
thank you

PR #330, opened 2019-04-23
remove Qt framework forcing on CMAKE_PREFIX_PATH

ra1nb0w:
I don't understand why it is fixed and here


f4exb:
Looking at history it was contributed a long time ago for Mac OS compatibility. It probably does not hurt but considering it refers to Qt 5.7 it is probably obsolete.

ra1nb0w:
right.

PR #358, opened 2019-06-02
Improve precision of master timer

Vort:
Changing of master timer type to `Qt::PreciseTimer` makes spectrum animation more smooth on Windows platform.

Here is the video with the same record being played.
On the left is original 4.8.2 version, on the right - with this patch applied.
[sdrangel_master_timer_precision.zip](https://github.com/f4exb/sdrangel/files/3244447/sdrangel_master_timer_precision.zip)


f4exb:
Makes sense. However will have to watch if it has an impact on CPU %.

f4exb:
No significative influence.

Issue #396, opened 2019-07-31
Request: documentation for developers


jmi2k:
Issue addressing the same problem as #98, but not only for demodulator plugins, but in general for the whole project. Some notes about how to start hacking some plugins, an overview of the structure of SDRangel and maybe one completely documented simple plugin to take as an example would make contributing easier.

They don't have to be very detailed nor very specific, just some pointers to where to start and some advice through the basics of writing plugins, and some notes about the overall architecture of SDRangel.

rgetz:
Like any open source project - documentation pull requests are always encouraged. :)

The easiest way is to start looking (or at least how I did it) was to run doxygen on the source to generate all the class listings, and call graphs. Its still missing doc, but the pictures, and understanding what calls what will help alot.

f4exb:
I am an advocate of Agile at work and in my personal projects also. One of the core four values of Agile is : "working software over comprehensive documentation". I think is says it all...

rgetz:
Having a few autogenerated pictures like:

![image](https://user-images.githubusercontent.com/1017869/62248635-9531cf00-b3b6-11e9-88d9-bf07304de491.png)

or

![image](https://user-images.githubusercontent.com/1017869/62249826-5cdfc000-b3b9-11e9-9519-1d93aa0ae579.png)


and allowing people to browse them (on http://f4exb.github.io/sdrangel/index.html or somewhere else) where they are all clickable - might be worthwhile (or it might generate a bunch of questions you don't want to answer - could go either way) On other projects I'm involved in - we have github pages updated from travis-ci on every release tag. I can do the same here, and send a pull request if you want. Most of the scripts should be exactly the same, you would just need to encrypt a github token that allows a travis-ci build to write into the repo...

Right now, I do run doxygen locally, and it does help from (at least my) newbie perspective. 

I can provide the Doxyfile if you want, so people can generate things themselves in a standard manner. And 5 lines of text in the wiki is also fine - your call. 

-Robin

jmi2k:
That's something, and indeed very useful, although I still think some proper documentation somewhere would foster contributions. But I understand that's something you would want to delay until the project is more stable.

@rgetz I'd appreciate anything which could help me get started writing a plugin. I haven't used Doxygen before

f4exb:
[...]

PR #398, opened 2019-08-02
add make target of 'dev_docs' which uses Doxygen to build html pages

rgetz:
This is not done by default, you need to manually type "make dev_docs"
in the configured build directory, as it takes ~6-8 min to build
everything.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -448,6 +448,45 @@ include_directories(
     ${OPENGL_INCLUDE_DIR}
 )
 
+# Set up optional make target for developer doc
+find_package(Doxygen)
+if(DOXYGEN_FOUND)
+  option(WITH_DOC "Generate developer doc with Doxygen" ON)
+  message(STATUS "added optional 'make dev_docs' for developer documentation generation")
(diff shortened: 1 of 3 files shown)


rgetz:
For those what want to have a look at what it creates, checkout :

https://rgetz.github.io/sdrangel/

the most interesting thing is the name spaces, and classes, and search. click on the side arrow to see the pictures, and call/caller structures.

I don't know if I would actually do this here (in this repo) - it does add 3.7G of html/svg files to the repo. I think just add it to the wiki, in the build section, should be fine more most developers.

This should close #396  (for now).

rgetz:
Just as a note, Eventually - I will remove my fork, and the on-line doc will be removed (I always think it’s bad form to have stale forks lurking). 

If it is OK - I will edit the wiki page https://github.com/f4exb/sdrangel/wiki/Compile-from-source-in-Linux to describe the host packages that need to be installed (doxygen + dot) and how long to wait.

f4exb:
I have added you as a collaborator which should enable you to update the wiki

PR #436, opened 2019-11-01
leansdr: fixes for macos

ra1nb0w:
- plscodes.LENGTH
  llvm < 3.9.0 doesn't support "Variable length array" and therefore
  can't be used (it is not part of c++11 standard). see commit
  https://github.com/llvm/llvm-project/commit/e7545b33ff79c55d08d78a35b052e494ea49a0bc

- float casting is needed to avoid confusione between double and float

- F_SETPIPE_SZ is not supported on macOS

### DIFF
--- a/plugins/channelrx/demoddatv/leansdr/dvbs2.h
+++ b/plugins/channelrx/demoddatv/leansdr/dvbs2.h
@@ -733,7 +733,7 @@ struct s2_frame_receiver : runnable
         // Read PLSCODE
 
         uint64_t plscode = 0;
-        complex<float> pls_symbols[plscodes.LENGTH];
+        complex<float> pls_symbols[s2_plscodes<T>::LENGTH];
         for (int s = 0; s < plscodes.LENGTH; ++s)
         {
             complex<float> p = interp_next(&ss) * agc_gain;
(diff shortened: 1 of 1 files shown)


plugins/channelrx/demoddatv/leansdr/dvbs2.h
```diff
@@ -733,7 +736,7 @@ struct s2_frame_receiver : runnable
         // Read PLSCODE
 
         uint64_t plscode = 0;
-        complex<float> pls_symbols[plscodes.LENGTH];
+        complex<float> pls_symbols[PLSCODES_LENGTH];
```
f4exb: Alternatively you could code this that I think would work (at least it does with clang while g++ lets it go...):
```c++
complex<float> pls_symbols[plscodes<T>::LENGTH]
```
... and avoids having to define a `PLSCODES_LENGTH`. A similar construct exists at line 224

f4exb: Alternatively you could code this:
```c++
complex<float> pls_symbols[s2_plscodes<T>::LENGTH];
```
and this avoids having to define `PLSCODES_LENGTH`. I tried it using clang and it fixes the variable length array issue.

ra1nb0w: oh, right!

PR #442, opened 2019-11-17
Virtual vs void

rsumner868:
There are quite a few mistakes in the original article and I don't think they should be any


f4exb:
Please do pull requests from same branches and not dev -> master. dev is devevlopment and still subject to changes. Most of your changes are already under way.

sdrbase/dsp/dspdevicemimoengine.cpp
```diff
@@ -311,7 +311,7 @@ void DSPDeviceMIMOEngine::workSampleSourceFifos()
     }
 
     std::vector<SampleVector::iterator> vbegin;
-    std::vector<SampleVector> data = sampleFifo->getData();
+    std::vector<SampleVector>& data = sampleFifo->getData();
```
f4exb: Already spotted. That was a long lasting bug just for a single character!

Issue #447, opened 2019-11-23
importing bookmarks does not work if there's no bookmark already


Antanide:
Linux, git commit acd0892 (v 4.12.1).

Fresh installation with no bookmarks.

You press the "import bookmarks" button and nothing happens.
Workaround: create a random temporary bookmark. Then the button works.

f4exb:
What is that you do not understand in "Import preset from file in the current group". Then you need to have a group defined (of course).

Useless to keep this issue opened.

PR #448, opened 2019-11-23
fix importing of presets when none is set

febs:
This should fix issue #447


f4exb:
Rejected as a breaking change

PR #480, opened 2020-02-19
Fixing CMake PkgConfig for the SoapySDR Module 

luigifreitas:
This Pull Request adds a case-sensitive alternative to the CMake's `pkg_check_modules` function (`SoapySDR` and `soapysdr`). The current code couldn't find the current version of SoapySDR automatically.

### DIFF
--- a/cmake/Modules/FindSoapySDR.cmake
+++ b/cmake/Modules/FindSoapySDR.cmake
@@ -9,7 +9,7 @@ if(NOT SOAPYSDR_FOUND)
     set(${VERSION} "${CMAKE_MATCH_1}" PARENT_SCOPE)
   endfunction(_SOAPY_SDR_GET_ABI_VERSION)
 
-  pkg_check_modules (LIBSOAPYSDR_PKG soapysdr>=0.4.0)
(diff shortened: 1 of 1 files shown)


cmake/Modules/FindSoapySDR.cmake
```diff
@@ -9,7 +9,7 @@ if(NOT SOAPYSDR_FOUND)
     set(${VERSION} "${CMAKE_MATCH_1}" PARENT_SCOPE)
   endfunction(_SOAPY_SDR_GET_ABI_VERSION)
 
-  pkg_check_modules (LIBSOAPYSDR_PKG soapysdr>=0.4.0)
+  pkg_check_modules (LIBSOAPYSDR_PKG soapysdr>=0.4.0 SoapySDR>0.4.0)
```
f4exb: is there a reason for the greater than (>) on SoapySDR and the greater or equal (>=) on soapysdr. Shouldn't it be greater or equal for both?

luigifreitas: Nope, this was a type. I corrected it.

PR #516, opened 2020-05-01
Added more flexible build options for external libraries

exception13:
Changes:
* Fixed concurrent building (added sdrbase dependency on limesuite)
* Don't get and build existed system libs if set ENABLE_EXTERNAL_LIBRARIES to AUTO
* Force use external library if ENABLE_EXTERNAL_LIBRARIES set to AUTO and  ${LIBRARY_NAME}_EXTERNAL set to ON

Example:
`cmake -DFORCE_SSE41=ON -DENABLE_EXTERNAL_LIBRARIES=AUTO -DLIBMIRISDR_EXTERNAL=ON -DLIBDSDCC_EXTERNAL=ON
`

libmirisdr and libdsdcc forced to external, other libraries will be taken from the system if possible

**WARNING: NOT TESTED ON WINDOWS AND MACOS!!!**


f4exb:
Limesuite does not depend on sdrbase this is the opposite.

.appveyor.yml
```diff
@@ -70,7 +70,7 @@ for:
             libusb-1.0-0-dev libboost-all-dev libasound2-dev libopencv-dev libopencv-imgcodecs-dev \
             libxml2-dev bison flex ffmpeg  libpostproc-dev libavcodec-dev libavformat-dev \
             libopus-dev libcodec2-dev libairspy-dev libhackrf-dev \
-            libbladerf-dev libsoapysdr-dev libiio-dev \
+            libbladerf-dev libsoapysdr-dev libiio-dev libairspyhf-dev librtlsdr-dev \
```
f4exb, line 73: No: these libraries are built from source.

.travis.yml
```diff
@@ -59,6 +59,7 @@ addons:
       - libairspy-dev
       - libhackrf-dev
       - librtlsdr-dev
+      - libairspyhf-dev
```
f4exb, line 62: Same remark as above

.appveyor.yml
exception13: * Depends removed

exception13:
> Limesuite does not depend on sdrbase this is the opposite.
sdrbase use limerfe. On my PC concurent build (8 threads) with external limesuite failed (at the time of build srdbase there are no headers and libraries of limesuite)

exception13:
> Anyway it is intentional that critical dependent libraries are built from source so there will be no other options. By experience being dependent on distributed libraries brings instabilities and lack of flexibility for implementing new features.

AUTO is optional. I'm build on Debian Testing and most libraries have the latest versions and it's reduce build time.

f4exb:
I entirely disagree on the debian/control file that should remain as-is. It is my intention not to use the packaged libraries here. Moreover Debian testing is not supported the resulting .deb is for Ubuntu 18.04 only and provided via Appveyor.

The opposite of "Limesuite depends on sdrbase" is 'sdrbase depend on LimeSuite"so we agree on this.

exception13:
* Reverted debian/control and CI scripts changes

PR #577, opened 2020-07-20
Set correct output directories for RelWithDebInfo build

Vort:
It can help to obtain precise results with performance profiler.


Vort:
Also I have made a commit, which is based on this change, it can be merged too: https://github.com/Vort/sdrangel/commit/8470fd89ab76074310f0d57cd9310bbe6ec7135e.
(it removes 3 divisions per sample in ATV demod, which results in ~10% increase of processing speed).

f4exb:
I should have performed the merge locally then push. Because it was done on Github and created an intermediate commit I had to cherry pick https://github.com/Vort/sdrangel/commit/8470fd89ab76074310f0d57cd9310bbe6ec7135e locally then push. Anyway it is now on master

PR #587, opened 2020-08-05
ATV Demod: introducing of new analog TV shader

Vort:
Here is my vision of how analog TV samples should be converted to digital pixels.

These commits introduce new class: `TVScreenAnalog`.
Which is capable of shifting each individual line with subsample precision.
With it comes `TVScreenAnalogData`, which contains image and line shifts data.
Since new screen now allows independent shifts, HSync algorithm is reworked.

New TV screen uses shaders to accomplish its tasks.
Because of this, it should be tested on various hardware.
As GPU drivers can contain bugs and execute shaders command wrong.

With separate test program > 1000 FPS was obtained with this shader.
Main time consumer in render function is texture copy command.
So shader is not needed to be optimized further right now.

CPU consumption should stay approximately at the same level.
Or may be even lower, because of possibility of data access inlining.

This code is not too much tested yet.
So it may be ready to merge as is or not.

### DIFF
--- a/plugins/channelrx/demodatv/atvdemod.h
+++ b/plugins/channelrx/demodatv/atvdemod.h
@@ -87,7 +87,7 @@ public:
     }
 
 	void setScopeSink(BasebandSampleSink* scopeSink) { m_basebandSink->setScopeSink(scopeSink); }
-    void setTVScreen(TVScreen *tvScreen) { m_basebandSink->setTVScreen(tvScreen); }; //!< set by the GUI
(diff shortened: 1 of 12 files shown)


f4exb:
AppVeyor fail due to 1 hour limit on MSVC build. Works fine on Linux. Not an issue.

f4exb:
As you mentioned it probably needs to be exercised a bit but I see no particular issues in merging it to mainstream. I had a quick look and it makes the code of the demodulator simpler as expected by moving the "pixel subsampling" hack away in a first class citizen rendering component which is nice.

Vort:
I have made additional tests and results looks good.
I know that several glitches are not fixed yet.
But do not want to overload this PR.

Issue #602, opened 2020-08-15
hamlib rigctl support?


hiperiondev:
Programs like Gpredict can correct on the fly frequency for Doppler effect on satellite communications. Usually use rigctl for this.
There is some support for this in sdgrangel?

f4exb:
No this is not the path taken in SDRangel. SDRangel can be fully controlled with its http REST API and this is the way to go to implement such things like Doppler shift control. In order to do this you have to write some intermediary software that will make the joint between data from Gpredict and SDRangel. This can be done in Python for example. There are examples in the `scriptsapi` folder and also in `swagger/sdrangel/examples`.

srcejon:
@hiperiondev, I've added a plugin that should support rigctrl in pull request #627 If you can try it out, it would be good to get some feedback from you.

f4exb:
This is a more general consideration but I start to like the idea of a "functional" plugin that this development makes emerging. I have only looked at the UI interface for the moment and I see that it is still pretty much embedded in the core (main) functions technically attached to the "Preferences" menu. 

Now what if this plugin would be able to generate several independent instances of itself like the channel plugins do? Then this would make it possible to control more than one device and one channel each instance opening its own port to receive commands. This is more in line with the spirit of SDRangel since its inception. And when I look at other items in the preferences menu the "LimeRFE control" is a good candidate also for a "functional" plugin, "AMBE" maybe also. It also makes it easier to manage dependencies if you just don't care about these functions and do not want to deal with SerialDV or LimeSuite dependencies. 

Similarly to what exists for channels then a canvas must be added to lay out the UI of these plugins and this adds even more stuff in the screen space which is pretty crowded already. However canvas can be stacked and this should not be a consideration that would prevent a development that could bring significant added value to the software.

For now I will keep it more or less like this as I need to put more thinking into this idea but some reorganization may take place in the future.

f4exb:
Two issues have to be addressed in the mid term:
  - `sdrbase/settings` should have no mention of `PluginManager`. The save/load of the RigCtrl plugin preferences should be handled in a different way
  - there is (or there should be) the possibility to retrieve the own instance API host address and port. These should be used with no possibility for the user to enter these parameters. This plugin should not be able to control other instances of SDRangel.

srcejon:
[...]

PR #627, opened 2020-09-08
Add rigctrl plugin.

srcejon:
This PR adds a plugin to support Hamlib's rigctrl protocol. This will allow software such as gpredict to adjust for doppler in SDRangel while tracking satellites. 

As per your suggestion in issue #602, I have used the existing REST API to implement this. i.e. the plugin opens a TCP port to receive rigctrl commands on, then translates them in to REST API commands. I see that you suggested this could be implemented in python, however, I think a plugin provides for a nicer/easier user-experience, as the plugin loads automatically and settings for the plugin can be managed in the GUI.

However, to support plugin settings in the GUI, I have made a couple of small additions to the plugin API (which you might not like, so happy to consider alternatives). I've added one callback to allow a plugin to add elements to the main window. In this instance, it allows the plugin to add a rigctrl menu item under the Preferences menu. Another callback allows these settings to be saved/restored along with SDRangels settings. I think this interface (or something like it) could be useful for other non-channel/device plugins, not just this one.

A description of how to test the plugin can be found in the readme.md.

plugins/misc/rigctrl - Add rigctrl plugin.
sdrbase/plugin/pluginapi.h/.cpp - Add misc plugin registration.
sdrbase/plugin/pluginmanager.h/.cpp - Add misc plugin registration.
sdrbase/plugin/plugininterface.h/.cpp - Add top level UI and global settings serialization callbacks.
sdrbase/settings/mainsettings.h/cpp - Allow plugins to save global settings in main settings file.
sdrgui/mainwindow.cpp - Load settings after plugins are loaded, to allow plugin settings to be loaded and saved.
sdrsrv/maincore.cpp - Support loading/saving of plugin settings.


f4exb:
I have to admit that the destiny of SDRangel is not solely in my own hands and I see that you are making significant contributions to this project so your voice counts. Then I have to admit also that including hamlib or other features like serial control of rigs, PTT, PA/LNA, filters, antennas switchover control, scanners and the like from SDRangel UI has become inevitable.  

This was indeed not my original intention and my suggestion was to use an external program that could collaborate with SDRangel and make the glue with data flow from other programs like gpredict using SDRangel http rest API. I am more into micro services possibly distributed than having a big centralized application.

However extending the plugin scheme to features other than channels and devices is interesting because it still leaves the choice to include or not this extra feature. Thus turning SDRangel into a big code bloat is not a fatality and remains a choice in the hands of the (savvy) user. In the same way you may want to exclude plugins for devices that you do not own or channel features you are not interested in (well currently there are no compile options to explicitly disable channel plugins from the build). 

I have to challenge the usage of the API for internal calls. Internal messaging could be used instead but this can be done at a later stage if the idea of "functional" plugins needs to be developed.

PR #638, opened 2020-09-18
Add modpacket channeltx plugin which implements a packet radio (AX.25/APRS) modulator

srcejon:
Hi Edouard,

The channeltx plugin adds support for a digital packet radio (AX.25/APRS) 1200bps AFSK modulator. Happy to change anything if needed. It's based on the NFMMod plugin. After addressing any feedback you might have, I'll try to look at adding 9600bps FSK support as well. It has been tested using Direwolf as the demodulator.


f4exb:
Looks very good! Thanks for the contribution. It is very complete and turn-key. There only a minor warning when compiling the `packetmodgui.ui` and perhaps to complete the REST API with all the settings. This would make the server variant fully functional which is quite interesting in the case of APRS because you may want to use a RPi in headless mode. 

All this can be done later before the next "official" release so this looks all good to merge for now.

srcejon:
Thanks! The settings in the web API that are missing shouldn't really be needed in normal use - they're mainly to help with debugging, but I can add them in. I'll also take a look at that warning.

f4exb:
It was mostly to pass the message parameters:
```c++
    QString m_callsign;
    QString m_to;
    QString m_via;
    QString m_data;
```

srcejon:
I added an action to support sending packets, that includes these. E.g: to send a packet from the command line:

curl -X POST "http://127.0.0.1:8091/sdrangel/deviceset/1/channel/0/actions" -d '{"channelType": "PacketMod",  "direction": 1, "PacketModActions": { "tx": { "callsign": "MYCALL", "to": "APRS", "via": "WIDE2-2", "data": ">Using SDRangel API to transmit" }}}'

f4exb:
Sure this can be done also and trigger output at the same tine.

PR #647, opened 2020-09-28
Increase supported sample rate in fileoutput and testsink by 10x

srcejon:
This PR increases the supported sample rates in the file output and test sink plugins by 10x from 9MSa/s to 90MSa/s. This is needed because I'm testing a modulator plugin that requires a sample rate greater than 9MSa/s.


f4exb:
And... does it really work at 90 MS/s ?

f4exb:
Thanks for your input 90 MS/s is indeed very fast so I would have been surprised that it would work with any classical hardware.

PR #651, opened 2020-10-02
Remove SDRGUI_API to allow linking on Windows

srcejon:
The current head fails to link on Windows for me, with:

webapiadaptersrv.obj : error LNK2001: unresolved external symbol "__declspec(dllimport) public: int __cdecl FeatureSet:
:getNumberOfFeatures(void)const " (__imp_?getNumberOfFeatures@FeatureSet@@QEBAHXZ) 

Removing SDRGUI_API from the class declaration fixes it.

(I'm not sure why your automated Windows build doesn't have the same problem too?)

### DIFF
--- a/sdrsrv/feature/featureset.h
+++ b/sdrsrv/feature/featureset.h
@@ -28,7 +28,7 @@ class Feature;
 class FeatureSetPreset;
 class WebAPIAdapterInterface;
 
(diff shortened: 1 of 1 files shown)


f4exb:
You should also remove #include "export.h"

Issue #617, opened 2020-08-26
USRP radio support via SoapyUHD


srcejon:
Ettus USRP radios (such as the B210) can be made to work with SDRangel on Linux via SoapyUHD.

Below, I try to build it in the same way as the other modules on: https://github.com/f4exb/sdrangel/wiki/Compile-from-source-in-Linux

Compile UHD from source, based on: https://files.ettus.com/manual/page_build_guide.html
```
sudo apt-get install libboost-all-dev libusb-1.0-0-dev python-mako doxygen python-docutils cmake build-essential
cd /opt/build
git clone https://github.com/EttusResearch/uhd.git
cd uhd/host
mkdir build
cd build
cmake -DCMAKE_INSTALL_PREFIX=/opt/install/uhd ../
make 
make install
/opt/install/uhd/lib/uhd/utils/uhd_images_downloader.py
# The following aren't required if installed to /
echo /opt/install/uhd/lib | sudo dd of=/etc/ld.so.conf.d/uhd.conf
sudo ldconfig
export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/opt/install/uhd/lib/pkgconfig
# Enable USB access from user accounts
cd /opt/install/uhd/lib/uhd/utils
sudo cp uhd-usrp.rules /etc/udev/rules.d/
sudo udevadm control --reload-rules
sudo udevadm trigger
```

Build SoapyUHD:
```
cd /opt/build
git clone https://github.com/pothosware/SoapyUHD.git
cd SoapyUHD
mkdir build; cd build
cmake -DCMAKE_INSTALL_PREFIX=/opt/install/SoapySDR ..
make
make install
```

With that, I have the USRP USB B210 channel 0 working for RX and TX in SDRangel 👍 

Channel 1 fails with:
terminate called after throwing an instance of 'uhd::runtime_error'
  what():  RuntimeError: Invalid recv stream command - stream now on multiple channels in a single streamer will fail to time align.

Most of the device specific settings seem to work. E.g. AGC, bus format, SR etc. One strange thing is when clearing the 'Auto DC corr' checkbox, the 'Global gain' is set to 0, ignoring what value it is set to. It gets the correct value when the checkbox is set though.

srcejon:
USRP radios can also work on Windows via Soapy. Instructions for getting that working are here: https://groups.io/g/sdrangel/message/1473

srcejon:
There are a few other issues:

[...]

PR #652, opened 2020-10-03
Add USRP radio source & sink plugins

srcejon:
This PR adds USRP sample source and sink plugins, adding support for these https://www.ettus.com/products/ SDRs.
Tested with a B210 on Windows and Linux, but should work with other devices. Seems to work better than the SoapySDR plugin, which crashes a lot. It requires the UHD library.


srcejon:
I've added the UHD library to sdrangel-windows-library project.
On Linux, UHD can be built as described here: https://github.com/f4exb/sdrangel/issues/617 and then when running cmake pass -DUHD_DIR=/opt/install/uhd

srcejon:
It seems the Linux build is failing due to a missing Python module called mako.

According to UHD build instructions, https://files.ettus.com/manual/page_build_guide.html, this is just required at build time.

This can be installed with 'sudo apt install python3-pip' then 'pip3 install mako'

Might also need:

pip3 install requests 
pip3 install numpy

The Windows build needs: https://github.com/f4exb/sdrangel-windows-libraries/pulls

f4exb:
It's a pity that some hardware support libraries have dependencies that have nothing to do with the actual interface. 

However adding python-mako and maybe python-cheetah to the list here: https://github.com/f4exb/sdrangel/blob/master/.appveyor.yml#L67 should suffice. 

For Windows it may indeed be easier to use pre-built binaries.

Alternatively the UHD support can be excluded from package builds and available only if compiled from source.

srcejon:
Thanks, I'll take a look.

For Windows, I have added the binary release to sdrangel-windows-libraries

srcejon:
Sorry, a bit of trial-and-error there. It works, but very close to 1h.

I see Debian actually has a pre-built UHD package that just can be installed with 'sudo apt-get install libuhd-dev'. That would speed it up, but I don't know how you prefer to manage the dependencies.

f4exb:
Yes for the package build it could be better to use the Debian package libuhd-dev.

PR #670, opened 2020-10-26
Usrp improvements

srcejon:
This PR includes a variety of improvements for the USRP plugins. Mainly:

Add LO offset option.
Work around bug in UHD library resulting in large TX LO leakage for low LPF bandwidths.
Check for reference and LO lock before streaming.
Set bit size to be 16 for interpolators and decimators.
Increase RX FIFO size, to reduce overflows at high sample rates.
Allow RX to continue streaming after receiving timeout.
Display master clock rate in GUI.

### DIFF
--- a/devices/usrp/deviceusrp.cpp
+++ b/devices/usrp/deviceusrp.cpp
@@ -22,6 +22,7 @@
 #include <cstring>
 #include <cmath>
 #include <regex>
+#include <thread>
 
 #include <uhd/types/device_addr.hpp>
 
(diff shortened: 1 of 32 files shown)


f4exb:
The Linux build uses `libuhd-dev` which is version 3.15.0.0 in Ubuntu 20.04 which is the image used. See: https://packages.ubuntu.com/search?suite=focal&keywords=libuhd-dev This may explain the error on the Linux build.

srcejon:
Ok, thanks. I'll see if I can get it to build with both versions.

Issue #660, opened 2020-10-16
Waterfall on Windows can be missing


srcejon:
I've seen a few cases where the waterfall can be missing on Windows (once in a VM while release testing due to lack of Open GL support in graphics driver, and another while running SDRangel via Remote Desktop Protocol, where the waterfall appears when run locally, but not remotely).

Both cases appear to be fixed by installing Qt's ANGLE libraries (libEGL.dll et al), which maps OpenGL to DirectX, if full OpenGL support isn't available.

I've a patch for this, which I'll open a PR for shortly.

f4exb:
I thought Qt's ANGLE was coming in by default but I may have been mistaken

zildjiani:
73 F4EXB j'ai un probleme avec sdrangel sdrangel-4.20.0-win64.exe apres execution du programme 
la fenetre du spectrumanalyzer ne fonctionne pas:couleur blanc!
compiuter win10 familliale 64bits  processeur i3 6G ram

merci d'avance

flogger76:
Hello,
I have the same problem. I'v installed sdrangel-4.20.0-win64, and te effect is like on this picture:
![sdrangle1](https://user-images.githubusercontent.com/37769094/97205866-50166b80-17b8-11eb-916e-bdebf7d88b2a.jpg)

I can't see the waterfall, spectrum, there is problem with refreshing frames inside the main window of program and sometime the program makes crashes.
My system:

Windows 7 64bit
uP Intel Core i5-2500 CPU 3,3GHz
graphic card Intel HD Graphic 2000, driver Intel v9.17.10.4229

Can you help me please what I have wrong ?

73!
Adam SQ1GPR

Davidi74:
Yes, waterfall is missing for me too in windows. I had reported(#513) in an earlier version also that the ATV decoder didn't refresh either, but the dev was unable to find the problem. Both issues occur on my two machines Which are windows 10 LTSC 2019

Davidi74:
Looks like this issue still persists on my Intel Iris based laptop with v4.21.0
Everything works fine on an my Nvidia 2080 based desktop.

I uninstalled the program on the intel iris based machine and found that the problem occurs with previous versions, however in  version 4.14.3 it works fine. So between 4.14.3 and 4.14.20 something appears to have changed.

f4exb:
Normally this is for Windows 10 not Windows 7 so anything can happen there. Anyway there is no much more that can be done.

Davidi74:
[...]

PR #671, opened 2020-10-27
Add ADS-B demodulator and GS-232 rotator controller plugins

srcejon:
Hi,

This PR adds an ADS-B demodulator plugin, that can receive position / heading / speed information broadcast by aircraft and display it in a table, map and forward to an aggregator.
It also includes a GS-232 rotator controller plugin. This can be used manually to set an antenna's azimuth and elevation, or linked to the ADS-B plugin (and potentiallly other plugins in the future) to track aircraft.
As this now requires a few extra Qt libraries, I've changed the Windows build script to use DeployQt. This should hopefully help with some of the OpenGL issues on Windows (#660) as it includes ANGLE.
The My Position dialog has been update to include a station name and altitude, for use on the map and calculating Az/El.
In order for the map to display on Windows, you'll need to merge the OpenSLL libraries PR.


alphafox02:
Just FYI, and maybe it's something I'm doing wrong, but installing the latest release with the awesome new ADS-B feature had two small issues in my case. I'm assuming because UHD support was added (my b205 works!) has resulted in my BladeRF coming up listed twice with no way to tell what driver it's using. I'm only making an assumption that it's listed twice due to the BladeRF driver and then again with the UHD driver.

I'm also unable to get a map to display in Linux. It's just a white blank area where the map should be even after aircraft are detected. Assuming again maybe an extra package is needed?

PR #673, opened 2020-10-28
RTL-SDR: Adjust sample rate limits

kasper93:


kasper93:
Set same limits as in librtlsdr https://github.com/osmocom/rtl-sdr/blob/ed0317e6a58c098874ac58b769cf2e609c18d9a5/include/rtl-sdr.h#L261-L262 

The original limit of 2.4MS/s is there because above that it _may_ be unstable. But it hardware dependent and we shouldn't limit this is software. I use 2.56MS/s in all other software and people use even more without any drops or other issues. Just want to get rid of this artificial limitation.

f4exb:
Makes sense. I have thought of doing this but lost track. Thanks for contributing!

kasper93:
@f4exb: What happened with this? It got merged into dev, but is lost now. It didn't work well?

f4exb:
True and I don't know what happened fortunately nothing gets lost with git. I will re-apply it.

PR #676, opened 2020-10-29
Replace qtDeviceName by alsaDeviceName in fcd* plugins

dl1ksv:
sdrangel works fine with the fcdpro and fcdpro+ devices if pulseaudio is used. 
But on installations with pure alsa and without pulseaudio these devices are detected but cannot be opened.
The failure looks like:

FCDProPlusInput::openFCDAudio: device with name FUNcube_Dongle_V2.0 not found

Switching from qtDeviceName to alsaDeviceName in openDevice() solves this issue. After this change sdrangel works flawless in a pulseaudio and an alsa environment

### DIFF
--- a/plugins/samplesource/fcdpro/fcdproinput.cpp
+++ b/plugins/samplesource/fcdpro/fcdproinput.cpp
@@ -85,14 +85,14 @@ bool FCDProInput::openDevice()
         return false;
     }
 
-    if (!openFCDAudio(fcd_traits<Pro>::qtDeviceName))
+    if (!openFCDAudio(fcd_traits<Pro>::alsaDeviceName))
     {
-        qCritical("FCDProInput::start: could not open FCD audio source");
+        qCritical("FCDProInput::start: could not open FCD audio source %s",fcd_traits<Pro>::alsaDeviceName);
         return false;
     }
     else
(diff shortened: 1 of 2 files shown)


f4exb:
Sorry but this is a breaking change. I have tried a diff of your PR with a FUNcube dongle and it does not produce any I/Q stream with your change.

dl1ksv:
Interesting. I tested on ubuntu 20.04 without pulseaudio and on gentoo with and without pulseaudio and it worked with both dongles in any cases.
After your observation I tried again , connecting / disconnecting and stopping and starting multiple times and suddenly I got no stream data.
But this did not only affect the fcd's but also my rtl-sdr. It seems that it depends on the start of the pulseaudio daemon.
Is pulseaudio already running or started by the client. Most of the time starting pulseaudio by the client resulted in no IQ data. Stopping and restarting SDRangel mostly solved the problem. But it's not reliable enough, to find the reason and propose a solution. 
I'll keep on searching.

kasper93:
Anyhow this would for sure break Windows.

f4exb:
Yes indeed so I think we can close this. Pulseaudio in Linux is a must anyway.

PR #678, opened 2020-10-30
Fix missing cmath include

zuckschwerdt:
This adds the `cmath` include to rigctlserverworker.cpp so the `std::abs()` on double overload is defined. Otherwise I get
```
plugins/feature/rigctlserver/rigctlserverworker.cpp:361:27: error: call to 'abs' is ambiguous
        bool outOfRange = std::abs(freq - targetFrequency) > m_settings.m_maxFrequencyOffset;
                          ^~~~~~~~
/usr/include/stdlib.h:132:6: note: candidate function
int      abs(int) ;
         ^
```

### DIFF
--- a/plugins/feature/rigctlserver/rigctlserverworker.cpp
+++ b/plugins/feature/rigctlserver/rigctlserverworker.cpp
@@ -22,6 +22,8 @@
 #include <QEventLoop>
 #include <QTimer>
 
+#include <cmath>
+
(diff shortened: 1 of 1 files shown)


f4exb:
`std::abs` is defined in `<cmath>` since c++17 however requirements are only for c++14. This does not seem to hurt anyway.

PR #682, opened 2020-11-04
NFMDemodSink: Proper audio scaling to 16-bit

kasper93:


plugins/channelrx/demodnfm/nfmdemodsink.cpp
```diff
@@ -258,11 +262,6 @@ void NFMDemodSink::applySettings(const NFMDemodSettings& settings, bool force)
         m_interpolatorDistance =  (Real) m_channelSampleRate / (Real) m_audioSampleRate;
     }
 
-    if ((settings.m_fmDeviation != m_settings.m_fmDeviation) || force)
```
f4exb, line 261: Not sure this is a good idea to get rid of expected FM deviation. I'll test that...

plugins/channelrx/demodnfm/nfmdemodsink.cpp
```diff
@@ -174,9 +176,11 @@ void NFMDemodSink::processOneSample(Complex &ci)
         if (!m_settings.m_audioMute && (m_settings.m_ctcssOn && m_ctcssIndexSelected == ctcssIndex || m_ctcssIndexSelected == 0))
         {
             Real audioSample = m_squelchDelayLine.readBack(m_squelchGate);
+            Q_ASSERT(audioSample >= -1.0f && audioSample <= 1.0f);
```
f4exb, line 179: Please do not add code that will make the whole app crash because of a particular audio sample.

kasper93: This will trigger only in debug build. And it is good reminder that we should work with values [-1, 1] at this stage. But if you think it is too much  I will remove it.

plugins/channelrx/demodnfm/nfmdemodsink.cpp
kasper93: Frankly speaking I'm looking forward for your input. We use `phaseDiscriminatorDelta` which always outputs in -1,1 range, without scaling, hence I removed it. Before we were multiplying signal my seemingly random values to boost it enough to be somehow in int16 range. Because proper scaling was missing.

As I understand it after demodulation we have floating point audioSample which by design should have -1,1 range. we can fixup this with scaling, but in fact it made things worse, because the range of signal was unexpected.

f4exb: Your change completely ignores the fact that along with a certain bandwidth comes a certain FM deviation. At least that's what most systems do. Anything larger than about +/-2.5 kHz is saturating. This is not correct. Try for yourself with a NFM transmitter and you will see. The present code has possibly room for improvement but this is not by dumping code that somehow works.

plugins/channelrx/demodnfm/nfmdemodsink.cpp
f4exb: Even in debug mode you would not like to be polluted by a problem that you are not currently investigating moreover if I'm right there is a clamping a bit further down.

f4exb:
Cherry picked:
  - Define _USE_MATH_DEFINES and clean M_PI redefinitions: 1e5ae8ea0f45a8041ab62c610f20363d3615ef80
  - NFMDemodSink: Proper audio scaling to 16-bit: 00ba5115f3a0b4d792ef1e7f397cfe5c51b7214f
  - NFMDemodSink: Fix ctcssOn logic: 7ad8c046f257bcac2866b9ebfec502befdfc79eb

Restored FM deviation and removed `Q_ASSERT`:
  - NFM demod: restored expected FM deviation: 38755e35f1915a59ecc41eb9fde4033b5318d7cc

plugins/channelrx/demodnfm/nfmdemodsink.cpp
kasper93: > Your change completely ignores the fact that along with a certain bandwidth comes a certain FM deviation. 

I mean what is the logic in current code apart from divining `output audio sample rate` by `freq deviation` which is big number which will just boost the signal. Which was probably done historically, because the scaling to 16bit was missing so other things ended up being used to amplify the signal. I might be missing something, but I completely don't see logic behind this value.

What I like without scaling is that whole signal including noise is in predicable range and we are not digitally clipping the signal later.

That say, this scaling factor (`setFMScaling`) should be something along the lines `1. / (2. * pi * m)` where `m` is modulation index, which for NBFM is << 1. Considering radio with maximum 3.5kHz audio and 2.5kHz deviation, we have m == 0.7.

Either way it would scale down the signal. But well it is linear scaling so I don't get the point of applying it in practice. We also have audio volume control to adjust that if needed.

Thanks for merging, I'm saying above, because I am considered about clipping that might happen now, but signal after demodulation and filtering should be quite low in practice so we dodge the bullet here, but hey this is the case of `dumping code that somehow works` :)

I want to try adding IIR filter before demodulation for frequency deviation and see how well it works.

f4exb: A scaling of 1,0 corresponds to a signal that goes all the way over the bandwidth implied by the sample rate. That is 48 kHz for 48 kS/s (this is complex) and in this case the audio produces will go from -1.0 to 1.0 so be full scale. If the deviation as per the modulation standard based on Carson law is say +/- 5 kHz corresponding to 16F3 (16 kHz max 3 kHz audio) then the audio will go from -5/24 to 5/24 i.e about [-0.2, 0.2] and this is not what we want so we need to scale up by 
4.8 (48/10 or 24/5). Experimentally this is a bit loud because the Carson law is not exact so I applied a -3dB (divide by 2) to this factor. This is why scaling to the expected FM deviation should be applied.

Although somehow working the existing code was flawed because of the improper normalization of filters and this scaling factor had to be reworked in order to meet audible signals. There is still this -3 dB but it make more sense.

To this occasion I also reviewed the bandwidths and deviations and made them more applicable to standards most prominently the 11F3 and 16F3 that are widely used.

PR #683, opened 2020-11-04
Add support for VS2019 and Ninja

kasper93:


kasper93:
First commit will be removed once those PRs are merged:
https://github.com/f4exb/sdrangel-windows-libraries/pull/7
https://github.com/szechyjs/mbelib/pull/26

It works fine, I tried to minimize changes to get it working. There are a lot that can be done to cmake files in order to make maintaining this build-system less pain. But maybe later, once there is working base (this) merged.

CMakeLists.txt
```diff
@@ -1,10 +1,6 @@
-cmake_minimum_required(VERSION 3.1.0)
+cmake_minimum_required(VERSION 3.16.0)
```
f4exb, line 1: This is creating a big mess probably unnecessarily. On Debian 10 cmake is at version 3.13.4 for example.

sdrbase/CMakeLists.txt
```diff
@@ -381,6 +380,7 @@ target_link_libraries(sdrbase
     ${sdrbase_FFTW3F_LIB}
     ${sdrbase_SERIALDV_LIB}
     ${sdrbase_LIMERFE_LIB}
+    Boost::headers
```
f4exb, line 383: Not working in many linux distributions that do not have the proper version of cmake

srcejon:
Nice patch - much faster building on Windows :)

For anyone else wondering how to build following this patch, I did:


Install Boost and adjust path below as needed, as no longer part of sdrangel-windows-libraries
Install Qt 5.15.1
Install ninja (https://ninja-build.org/)
Run VS2019 shell  
```
git clone https://github.com/f4exb/sdrangel.git
cd sdrangel
git submodule update --init --recursive
mkdir build
cd build
cmake -Wno-dev -G Ninja -DCMAKE_BUILD_TYPE=Release -DFORCE_SSE41=ON -DDEBUG_OUTPUT=ON  -DRX_SAMPLE_24BIT=ON -DCMAKE_PREFIX_PATH=C:\\Qt\\5.15.1\\msvc2019_64;C:\\Libraries\\boost_1_73_0 ..
ninja        
ninja package  
```

kasper93:
>  Install ninja (https://ninja-build.org/)

You can use one for  VS Installer, it is somewhere in components to select.

> ninja
> ninja package

No need to do this twice, just call ninja once with target. And I would suggest `ninja install` with `-DCMAKE_INSTALL_PREFIX=...` passed to cmake.

PR #686, opened 2020-11-06
CI: Add docker build

kasper93:


kasper93:
Just a simple build to cross validate docker build images. Needs this to work with PRs https://github.com/f4exb/sdrangel-docker/pull/25 

Up to your decision if you want this build, it was simple to add.

f4exb:
Looks like a good idea. I'll try to see why it fails currently.

f4exb:
I know what the issue. It tries to find this commit https://github.com/kasper93/sdrangel/commit/a21c94c355e13a84d93fb322430afbdc3f4af2b1 but this is in your fork not in mainstream so it can't find it.

kasper93:
Yes, that's why I added fetch to pull commits form PR. 

But, sorry. I messed up, pushed wrong thing to the other repo. 

It needs to be
```
git fetch origin ${sdrangel_tag}
```
instead...

I tested it locally, don't know how I lost `origin` part.

f4exb:
The issue why the Docker build was hanging is actually fixed with `APPVEYOR_CONSOLE_DISABLE_PTY: true`. Not sure now BYOC is needed to overcome job time limit however now it is very fast on my 8 CPU Xeon server (named "nasrock").

kasper93:
> to overcome job time limit

We would need to preserve docker cache between build. Most layers doesn't need to be rebuild. And main part would get updated because we have commit hash as argument.

```
# save
docker save sdrangel:latest | lz4 -9 - docker_cache.lz4
# restore
unlz4 -c docker_cache.lz4 | docker load
```

And cache this `docker_cache.lz4` in appveroy between builds.

PR #688, opened 2020-11-06
ADS-B demodulator improvements

srcejon:
Add OpenSky Network aircraft database support, for information about aircraft model, owner, registration.
Add airline logos and country & military flags for display in the table.
Add OurAirports airport database support, to allow airports and ATC frequencies to be displayed on the map.
Allow ATC frequency to be tuned by clicking on the map.
Add support for displaying flight paths on the map.
Allow columns in table to be rearranged and hidden.
Allow rows in table to be sorted by clicking on header.
Allow switching units from ft, kn, ft/min to m, kph, m/s
Allow aircraft timeout to be set by the user.
Allow font used for the table to be set by the user.
Add optional display of demodulator statistics.
Support multithreading in demodulator to reduce FIFO overflows.
Add support for demodulating all Mode-S frames and feeding them.
Add support for feeding in Beast hex format.
Allow option of correlating against full preamble or partial preamble.
Supporting highlighting of an aircraft in the table by selecting it on the map.
Use difference of zeros and ones sums, rather than absolute threshold, to better account for varying conditions and make the threshold easier to set.
Enable anti-aliasing for text on the map.
Improve CRC performance by 5x-10x.
Move map resources to plugin from sdrgui.


sdrbase/util/crc.h
```diff
@@ -105,6 +114,8 @@ class SDRBASE_API crc32 : public crc
     crc32() : crc(32, 0x04C11DB7, false, 0xffffffff, 0xffffffff) {}
 };
 
+// Should probably try to use SSE 4.2's CRC32C instruction if available
+// See _mm_crc32_u8 and _mm_crc32_u64 intrinsics
```
f4exb, line 118: Problem is that it should also compile for ARM

srcejon: Sure. It doesn't currently use them, and if it did, it would be implemented so that there would be an alternative if not available. 
Some versions of ARM processors also have crc32c insn, but not all.

f4exb: BTW when I say "ARM" I mean armv8 a.k.a. arm64. I do not aim for support of earlier versions. Practically this addresses RPi 3B+ and RPi 4

f4exb:
I just tried it and don't get me wrong for what follows I like it a lot and you made a hell of a big and great job! So here are just a few things that in my opinion need scrutiny:
  - The airport and aircraft downloads are unconditional even if you have already downloaded the data. Maybe there could be a confirmation.
  - There should be the ability to activate or de-activate the VHF radio link because if you run only one device set then by default it will tune the current receiver that is precisely on the ADS-B frequency effectively de-tuning it. Indeed this is only by actively clicking on the VHF frequency but still... if you are like me wanting to play with all buttons... :smile:
  - The columns of the table do not resize with data contents resulting in a few ellipsis and then you need to resize them by hand. There could be an option (push button) in the settings to force resize.
  - After changes to the threshold mechanism It is hard for me to find the right threshold to optimize reception and relate it to what I see in the spectrum. Why the range starts only at -45 dB now? Anyway I haven't look at the details yet.

f4exb:
I think I know why this threshold setting is a bit weird. This is a difference expressed in dB and in fact judging by its value it is very minute so as to get some decodes. At least in my experience you need to be below -30dB that is less than 0.001. In fact it does not make sense to express a difference in dB because dB is a log and logs are suitable for ratios. As a matter of fact I am wondering if the ratio of ones vs zeros wouldn't be a better metric. I will experiment with that.

srcejon:
Don't worry, feedback is welcome :)

- Yep, confirmation sounds reasonable. I'll add it.
- Perhaps for this, it might make sense if the list of devices was a bit more selective. E.g. we could exclude the device the ADS-B demod is running on, unless it has an AM Demod running on it as well (i.e. for someone with only one SDR who does actually want to switch between ADS-B and ATC). A disable button is pretty trivial to add as well, so I can add that too.
- Sure, I'll add an auto-resize option.

The correlation may still need some further improvements, as I've only spent a little bit of time on it, but I can try to explain why I changed it. Most significantly, I was failing to receive frames that could be received previously (i.e. from an .sqriq file), regardless of what threshold was used. As you'd implemented it, it appeared to require the received signal to have very specific power levels. E.g. the ones needed to be above it, and the zeros below it. 
The problem I was finding with this, is that there isn't one threshold that you can set that will work in a wide range of scenarios. E.g: Lets say you want to be able to receive a weak signal from a plane far away: you'd set the threshold just above the noise floor, and this would indeed work for weak signals. However, the problem with this, is that it could then fail to receive a frame from an aircraft with a very strong signal that was directly overhead. It could just be that the noise floor has moved up (exceeding the threshold for the zeros), or if you happen to live near a busy airport, you may receive multiple frames from different aircraft simultaneously. A strong signal should still be able to be received over a weak one, but if you have a threshold for the zero chips that you need to be under - that probably will not be the case if there is another weak signal in the background. If you then set the threshold higher to combat this, you'll then fail to receive weaker signals. 
Another problem with manually setting absolute power thresholds is that it doesn't really work with AGC as the noise floor can move around (at least not unless the demod knows what the AGC is doing, but that isn't the case for us, as we don't get this information from the SDRs). 
For me, it appeared to be the case, that if I set the threshold too low, I wouldn't receive anything at all. With what I've now implemented, setting the threshold too low increases your chance of decoding something that isn't actually a real signal, but it shouldn't miss frames.
Fundamentally, I don't think the correlation should be based on the absolute power level. It's should just try to see if there is something that appears to be modulated, regardless of the absolute power levels. 
For tuning the threshold, I'd suggest starting at -20dB, then looking at the minimum and average values shown in the table. If the minimum values are the same value as the threshold, you can try lowering it. If above, you can increase it, and reduce the likelihood of decoding an invalid frame. Note if you enable displaying the statistics, you can see a count of the number of times the correlator incorrectly triggers and also the effect on processing time.
Starting the range at -45dB is purely empirical based on what I'd seen when I'd tried it with the USRP, Lime and RTL SDRs I have here, with a couple of different antennas and external LNAs. I wasn't finding I'd need to go anywhere near that low, but thought I'd set it well below what I'd needed, in case others did.
If it is not working well for you, please feel free to send me some .sdriq files so I can take a look. I'm sure there are situations that I haven't considered and that it can be improved. It would be useful to build up a test set of .sdriq files from different environments, so any changes to the algorithm can be properly benchmarked.

srcejon:
It's interesting that you say you need a threshold of -30dB. Which SDR is that with and what gain? I will double check, but for my Lime, USRP and RTL SDR, I think I was using around -25 to -15. 
Yes, I left it in dB, because I thought numbers like -30 to -10, would be easier to use than a dial that went from 0.1 to 0.001. However, a different metric may indeed make more sense.

f4exb:
That's a Lime Mini

f4exb:
Back on the threshold thing I completely agree that with my initial change the validation of the sync sequence was entirely based on a stable noise floor and for good and bad reasons this is not always true. It happens that my tests were done in a quiet environment with only a few frames at night time and as you know air traffic is reduced globally these days...

First the bad reason for a moving noise floor: when working ADS-B at 1090 MHz you should put yourself in the situation where an out of band signal (GSM, TACAN pulses, ...) is not affecting your global gain. Good SDR practice starts with hard filtering. As ADS-B monitoring is popular it is easy to find good SMA mounted 1090 MHz band-pass filters these days.

The good reason: ADS-B is an Aloha kind of system that is every one speaks up without co-ordination and because messages are repeated there will be statistically a time when it will be clear enough to pick up the signal. This is still subject to the number of "speakers" but as you know in practice it works. Thus incoming frames may jump in and contribute to the "zero" sum significantly above the real noise floor. This is exactly what you noticed in a busier environment and why my approach was not robust.

Now all systems generally expect a certain SNR to be able to decode a signal and ADS-B is no exception. To estimate the SNR of a particular frame you only have the sync sequence to start with and bursts are so short that you may assume that the SNR will not vary too much for the rest of the message. So if you want to put a threshold to decide if a signal is good enough for you you should have it in terms of SNR and SNR is signal **over** noise **ratio**.

This is why I think you should work with a ratio rather than with a difference. This will make it clear and consistent over the wide range of signal powers. The rest of your demonstration is perfectly valid.

srcejon:
Sounds reasonable. Did you have something in mind? (I went for the difference simply because it was a quick change to your code that didn't add any extra processing, as you was concerned about the performance of the original version. However, I then added the multi-threading afterwards, so we should have some spare performance if needed)

srcejon:
On the subject of multithreading, I should probably rewrite the code to use message queues rather than sempahores.

f4exb:
> Sounds reasonable. Did you have something in mind? 

Yes I can take this part and experiment with it. It does not need extensive change it is just a different way to estimate the valid threshold. Also the correlation display may have to change a little bit to show more significant values.

> On the subject of multithreading, I should probably rewrite the code to use message queues rather than sempahores

Maybe... it depends on how reactive you may want this to be. Message queues have an overhead over direct semaphore processing.

PR #690, opened 2020-11-07
NFMDemod: Add RF filter for freq deviation

kasper93:


kasper93:
It is the last time I promise...

plugins/channelrx/demodnfm/nfmdemodsink.cpp
```diff
@@ -256,14 +269,16 @@ void NFMDemodSink::applySettings(const NFMDemodSettings& settings, bool force)
 
     if ((settings.m_rfBandwidth != m_settings.m_rfBandwidth) || force)
     {
-        m_interpolator.create(16, m_channelSampleRate, settings.m_rfBandwidth / 2.2);
-        m_interpolatorDistanceRemain = 0;
-        m_interpolatorDistance =  (Real) m_channelSampleRate / (Real) m_audioSampleRate;
+        m_interpolator.create(16, m_channelSampleRate, settings.m_fmDeviation);
```
f4exb, line 272: This does not respect Carson's rule where bandwifth = 2 * (peak_deviation + fmax). The deviation being used in the code is twice the peak deviation so this will account for the `/2` in the Interpolator polyphase filter that is real based. If you take a &plusmn; 5 kHz deviation (10kHz for "our" deviation) and 3 kHz max audio bandwidth this makes 2 * (5+3) = 16 kHz and not 20 kHz as it will in fact result. Not a big deal. For &plusmn; 2.5 kHz Carson's rule yields 11 kHz vs 10 kHz not a big deal either. So it will work for the current 11F3 and 16F3 standards but this is only by chance.

We could base filtering entirely on FM deviation, AF bandwidth and apply Carson's rule but I am afraid that this will make it a bit too complex to the average user and NFM is in fact quite popular.

For now there is a mix of Carson's rule and fixed setting for lowest bandwidth in `nfmdemodsettings.cpp` and I am wondering what's wrong with this:

```c++
//  fixed      |Carson (3k)                      |Carson (6k)
//             |      11F3   16F3                |
const int NFMDemodSettings::m_rfBW[] = {
    5000, 6250, 8330, 11000, 16000, 20000, 25000, 40000
};
const int NFMDemodSettings::m_fmDev[] = {
    2500, 2500, 3330, 5000,  10000, 14000, 19000, 28000
};
```

Note that if you use the REST API RF bandwidth and expected FM deviation are independent.

kasper93: > This does not respect Carson's rule where bandwifth = 2 * (peak_deviation + fmax).

I think I tried to apply deviation too eagerly, but before it was BW / 2.2 . The point here is for NBFM where modulation factor is <<1 bandwidth is approximately 2*audio bw. Full Carson's rule is applicable to WBFM where modulation factor is > 1.

If we agree on this we could remove whole BW selection and base everything on the audio bandwidth control. This would be easy for user to adjust one parameter instead of two.

f4exb: Carson's rule is what applies to decide for the 11F3 (11 kHz BW 3 kHz audio) and 16F3 (16 kHz BW 3 kHz audio) so this is not limited to wideband transmissions. The result of applying bandwidth as 2 times the peak deviation as in this change is that the assumed maximum audio bandwidth is the peak deviation. This is different from Carson's rule but this indeed makes some sense since the more deviation the more audio bandwidth you are able to carry without significant distortion.

I think this can be merged as is since there will be no significant difference to the average user who deals only with 11F3 and 16F3 basically. However this renders the RF bandwidth setting inoperative and this is annoying in the longer run.

I think that eventually the RF bandwidth can be restored as before and made operative again and in the GUI the 3 settings (RF BW, AF BW, FM dev) left to the (savvy) user control (same as with REST API) but there would be a choice of predefined standards acting immediately on the controls recovering the situation when you get lost. These standards may be related to the channel spacing (this is what matters in the end) and can follow [this rule] https://en.wikipedia.org/wiki/Frequency_deviation#:~:text=FM%20applications%20use%20peak%20deviations,spacing%20or%205%20kHz%20spacing) than encompasses the mostly used 12.5 and 25 kHz channel spacing (11F3 and 16F3 respectively). As one can see spacing less than 8.33 kHz cannot carry voice audio but this is expected. For spacing larger than 25 kHz there used to be a 36F3 standard which is a kind of a waste of bandwidth and could be made 36F6 instead.

Issue #496, opened 2020-03-29
Direct audio input/output sample device plugin


f4exb:
Following https://groups.io/g/sdrangel/topic/sdrangel_as_audio_spectrum/72473800 this ticket is opened to create an input and an output plugin to deal with audio stereo input and output as a sample source and sink plugin.

hiperiondev:
Some notice of this enhancement?

f4exb:
Output plugin done here: https://github.com/f4exb/sdrangel/commit/87d8816881d146d73a55368759af0c437ef0a8d2

f4exb:
Audio Output implemented in v6.1.0

PR #699, opened 2020-11-09
Add audio input sample source

srcejon:
This is a first pass of implementing an audio sample souce plugin (partially addressing #496).

It uses sdrbase/audio/audiodevicemanager. While it seems to work, there are a couple of improvements that could be made:

AudioDeviceManager::getInputDeviceIndex possibly needs to be updated to use the realm (https://doc.qt.io/qt-5/qaudiodeviceinfo.html#realm) if on qt >5.14, as on Windows, there can be duplicate devices with the same name. One set with default realm and the other with wasapi. 
It would also be useful if there were a way to re-enumerate the audio devices determined in AudioDeviceManager::AudioDeviceManager(), to account for a user attaching a USB audio device while SDRangel is running.
When only one of the L or R channels is being used to drive I, with Q=0, I wanted to set the main spectrum window to show only the real half of the spectrum. However, calling setSsbSpectrum didn't seem to work.
The plugin possibly shouldn't have been named AudioInput, as it clashes with existing sdrbase/audio/AudioInput class - but I thought consistent user-visible naming was probably more important. Feel free to change it though!


plugins/samplesource/audioinput/readme.md
```diff
@@ -0,0 +1,46 @@
...
+<h3>6: Volume</h3>
+
+A control to set the input volume. This is not supported by all input audio devices.
+
+<h3>7: Channel Map</h3>
+
+This controls how the left and right stereo audio channels map on to the IQ channels.
+
+* I=L, Q=0 - The left audio channel is driven to the I channel. The Q channel is set to 0.
+* I=R, Q=0 - The right audio channel is driven to the I channel. The Q channel is set to 0.
```
f4exb, line 44: I wouldn't leave this possibility. It is assumed that the IQ data stream is a truly complex stream and channels also assume this input so I would keep only the possibility to swap I/Q. If the user feeds only one channel and gets crap it is the user responsibility.

f4exb: Of course then the spectrum issue is no more an issue.

f4exb: Note that there is a possibility to use a real signal as input. Similarly to what is done in Pavel Demin's receiver for Red Pitaya (at a much higher sample rate) the real signal can be mixed with two NCOs in quadrature and then obtain an analytic complex signal centered around the NCO frequency. 

One can possibly use a complex NCO and have I = Q in input thus the two real NCOs in quadrature is in fact provided by a single complex NCO and the complex signal is directly obtained in output.

The NCO frequency can be set to half the bandwidth of the input signal.

f4exb: https://github.com/f4exb/sdrangel/issues/706 opened

PR #714, opened 2020-11-19
CMake: Make compiler flags selection and detection more flexible

kasper93:
- Fix and enable sdrbench build on Windows


CMakeLists.txt
```diff
@@ -281,45 +276,11 @@ set(CMAKE_INSTALL_NAME_DIR
...
-      add_compile_options(-fsanitize=address -fno-omit-frame-pointer -g)
-      add_link_options(-fsanitize=address)
-    endif()
-endif()
-
-# set compiler
-include(FindCompiler)
-
-if (C_CLANG OR C_GCC)
-	set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Wvla -Woverloaded-virtual -ffast-math -ftree-vectorize ${EXTRA_FLAGS}")
```
f4exb, line 312: Why remove this?

CMakeLists.txt
```diff
@@ -281,45 +276,11 @@ set(CMAKE_INSTALL_NAME_DIR
...
-      link_libraries(clang_rt.asan_dynamic-${ASAN_LIB_ARCH} clang_rt.asan_dynamic_runtime_thunk-${ASAN_LIB_ARCH})
-      add_link_options(/wholearchive:clang_rt.asan_dynamic_runtime_thunk-${ASAN_LIB_ARCH}.lib)
-    else()
-      add_compile_options(-fsanitize=address -fno-omit-frame-pointer -g)
-      add_link_options(-fsanitize=address)
-    endif()
-endif()
-
-# set compiler
-include(FindCompiler)
```
f4exb, line 309: Why did you remove cmake/Modules/FindCompiler.cmake. How can I tell between GCC and CLANG? This is important in other parts of the software!

CMakeLists.txt
kasper93: Not removed, moved to find flags cmake. `FindCompiler` was duplicated there and I stared this patch before you actually fixed this warnings. But same flags are set https://github.com/f4exb/sdrangel/pull/714/files#diff-1b53431f5e291976688daa17a2a78f1c0d4ca5bc0cd5559f9aed23844ba2789fR217

kasper93:
Seems like build env is busted on appveroy. I will retrigger build, because I have small change and if it fails again probably need to add libclang1 to installed pacakges, not sure why it desycnced. 

```
/usr/local/bin/doxygen: error while loading shared libraries: libclang-9.so.1: cannot open shared object file: No such file or directory
CMake Warning at /usr/share/cmake-3.16/Modules/FindDoxygen.cmake:447 (message):
  Unable to determine doxygen version: 127
Call Stack (most recent call first):
  /usr/share/cmake-3.16/Modules/FindDoxygen.cmake:610 (_Doxygen_find_doxygen)
  CMakeLists.txt:452 (find_package)
-- Found Doxygen: /usr/local/bin/doxygen (found version "") found components: doxygen dot 
/usr/local/bin/doxygen: error while loading shared libraries: libclang-9.so.1: cannot open shared object file: No such file or directory
CMake Error at /usr/share/cmake-3.16/Modules/FindDoxygen.cmake:689 (message):
  Unable to generate Doxyfile template: 127
Call Stack (most recent call first):
  CMakeLists.txt:452 (find_package)
-- Configuring incomplete, errors occurred!
```

CMakeLists.txt
kasper93: Because it was duplicated.

f4exb: Indeed I don't know why the code of FindCompiler landed in FindCPUFlags bet they could be kept separate as they do not relate logically to each other. And I never use the SANITIZE_ADDRESS so I don't mind removing it.

CMakeLists.txt
f4exb: It is still not clear where these flags are set now.

kasper93: Ok, I will change it to make more sense.

f4exb:
https://github.com/appveyor/ci/issues/3572 opened for Ubuntu 20.04 build but VS2019 still fails.

f4exb:
I think doxygen comes as a dependency of another pacakge. As suggested the Ubuntu build issue in Appveyor could be circumvented by adding this snippet after https://github.com/f4exb/sdrangel/blob/master/.appveyor.yml#L89
```yaml
- sh: |
	curl -fsSL -o doxygen.tar.gz "https://phoenixnap.dl.sourceforge.net/project/doxygen/rel-1.8.17/doxygen-1.8.17.linux.bin.tar.gz" && \
	tar -xzf doxygen.tar.gz && \
	sudo cp -a doxygen-1.8.17/bin/doxy* /usr/local/bin
```

kasper93:
I fixed build, rest of changes when I have more time at hand.

PR #715, opened 2020-11-20
CMake: Split arch detection and compiler flags into separate files

kasper93:


kasper93:
I split compiler flags, arch detection and compiler detection to separate files. To make it more intuitive what's what as you suggested.

kasper93:
And I removed compiler detection all together, because  what we need is MSVC-like compiler vs. gnu-like compiler and this is provided natively by cmake.

CMakeLists.txt
```diff
@@ -281,8 +281,8 @@ else()
     message(STATUS "Compiling for 16 bit Rx DSP chain")
 endif()
 
-# find cpu flags (and set compiler)
-include(FindCPUflags)
+# Set compiler options based on target architecture and selected extensions
+include(CompilerOptions)
```
f4exb, line 285: Much better. Thanks for the update.

cmake/Modules/CompilerOptions.cmake
```diff
@@ -1,13 +1,12 @@
 include_guard(GLOBAL)
 
-include(DetectCompiler)
```
f4exb: I still have to make the difference between GNU and CLANG because libsigmf is GNU specific. How does cmake resolves this?

f4exb:
Get core dump at startup https://github.com/f4exb/sdrangel/issues/716 refers

Issue #716, opened 2020-11-20
Core dump at startup while loading first plugin


f4exb:
Compiled fresh from master. Then:
```
2020-11-20 21:53:06.606 (D) MainWindow::MainWindow: load plugins...                                                                                                                                                                         
2020-11-20 21:53:06.606 (D) PluginManager::loadPluginsPart folder /opt/build/sdrangel/build/../lib/sdrangel/plugins is empty                                                                                                                
2020-11-20 21:53:06.607 (D) PluginManager::loadPluginsDir: fileName: libbeamsteeringcwmod.so                                                                                                                                                
Segmentation fault (core dumped)   
```

f4exb:
```
[Current thread is 1 (Thread 0x7f9830d3f000 (LWP 1272716))]
(gdb) bt
#0  std::__atomic_base<int>::load (__m=std::memory_order_relaxed, this=0x0) at /usr/include/c++/10/bits/atomic_base.h:426
#1  QAtomicOps<int>::loadRelaxed<int> (_q_value=...) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qatomic_cxx11.h:239
#2  0x00007f982c019af2 in QBasicAtomicInteger<int>::loadRelaxed (this=0x0) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qbasicatomic.h:107
#3  0x00007f982c01a74c in QtPrivate::RefCount::ref (this=0x0) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qrefcount.h:55
#4  0x00007f982c01a9f7 in QString::QString (this=<optimized out>, other=..., this=<optimized out>, other=...) at /usr/include/x86_64-linux-gnu/qt5/QtCore/qstring.h:1051
#5  0x00007f982c02320c in __static_initialization_and_destruction_0 (__initialize_p=1, __priority=65535) at /opt/build/sdrangel/plugins/channelmimo/beamsteeringcwmod/beamsteeringcwmodplugin.cpp:37
#6  0x00007f982c023343 in _GLOBAL__sub_I_beamsteeringcwmodplugin.cpp(void) () at /opt/build/sdrangel/plugins/channelmimo/beamsteeringcwmod/beamsteeringcwmodplugin.cpp:93
#7  0x00007f982c027eee in _sub_I_65535_0.0 () at /opt/build/sdrangel/sdrbase/dsp/channelmarker.h:28
[...]

PR #717, opened 2020-11-22
Fixes for clang-cl build

kasper93:


kasper93:
Originally I had only clang-cl changes, but added also few more to fix #716

f4exb:
There could be other adverse effects of LTO if the `volatile` keyword has been omitted in places and there is no guarantee that every legitimate volatile object has been properly flagged as such. However your implementation makes it easy to switch it off if some suspect behavior occurs.

PR #723, opened 2020-11-24
Add VOR demodulator

srcejon:
Hi Edouard,

This is a tad obscure, but a little bit of fun. VORs are VHF navigation aids used by aircraft. This plugin can receive the signal from multiple VORs in order to calculate your position, which is displayed on a map. More details in the readme.


f4exb:
Sounds fun. I am a bit obfuscated from Nice airport I'll try anyway 😄

f4exb:
Feature plugins are not yet quite up to this job but this is how I see VOR implementation in the future:
  - The VOR demodulator will take care of a single VOR just like the other demodulators or receiving channels
  - You can (pre) allocate several VOR channels randomly (well not so random, you have to think a little about it of course...) possibly in different device sets which opens multi receiver capability. This can be handy if you want to collate data from VORs at both ends of the band and have a pair of RTL-SDRs that cannot cover 10 MHz bandwidth individually
  - The VOR (coordinator) feature plugin will handle the map and orchestration that is today present in a single demodulator. It will pick up VOR channels present in the system (sensibly...) to accomplish the task. Each VOR channel will expose its bearing data in a data channel (yet to be designed so this is why I said feature plugins are not mature yet for this).

srcejon:
Sounds good. It would be useful if different features can send data (and control information) to each other as well. Thanks.

PR #746, opened 2021-01-13
Add PacketDemod, APRS, Map and Star Tracker plugins

srcejon:
Hi Edouard,

Here are four new plugins:

- PacketDemod to work with PacketMod (currently AFSK only)
- APRS Feature that displays APRS packets received by the PacketDemod, including sending symbols to the new Map feature for display and plotting charts for weather / telemetry
- Map feature for combing the display of items from other plugins, including APRS, ADS-B demod and Star Tracker
- Star Tracker plugin for radio astronomy and EME. Can track Sun, Moon or Stars and drive the Rotator Controller plugin, as well as interface with Stellarium

This also includes some updates to the GS-232 controller and ADS-B Demod to work with the new pipes and a few small extra features.

In order to support the above, I needed to change the messages pipes to support both ChannelAPI and Features as the source of messages (so I added a class PipeEndPoint that they both inherit from). These changes can be found in the first commit. If you think this should be done in a different way, please let me know.

I plan to do more work on these, but I thought I'd get some feedback, as the patch is already quite big!

To generate a passcode for the APRS-IS internet gateway, see: https://apps.magicbug.co.uk/passcode/


srcejon:
Yep - only the source currently, but it would potentially be useful to have the destination generic too - but I wanted to hear your thoughts before changing too much. 

E.g. The next step of the APRS client will be to transmit as well, so it will require some type of Feature to Channel messaging. That could be done with the Web API TX Action as well. One complication of using pipes would be how the user specifies which PacketMod to use (as you could have multiple on different frequencies). We could possibly have the pipes created with names that include channel index or something like, but not sure that's ideal. Any thoughts?

f4exb:
OK finally found what was wrong in the Docker build: 
```
fatal: ambiguous argument 'soapy-sdrplay-0.2.0': unknown revision or path not in the working tree
````
I think I stumbled upon this already. Some guys remove tags which is really bad practice and proves again that SoapySDR stuff is cr..p.

I'll fix that later.

PR #766, opened 2021-02-06
Few small fixes

kasper93:


kasper93:
I had those changes in my local branch and I figured I will upstream them. Nothing much, few minor changes here and there.

PR #797, opened 2021-03-05
Add Costas loop PLL to channel analyzer

srcejon:
This patch:

- Adds a Costas loop PLL to the channel analyzer
- Adds loop bandwidth and other PLL controls to Channel Analyzer GUI.
- Fixes bug where PLL lock frequency is incorrectly scaled by the decimation factor.

Costas loop should somewhat combine existing PLL and FLL functionality, providing both frequency and phase lock. Seems to provide a more stable lock too in the brief testing I've done.  
Happy to change the way the GUI hides the extra widgets when PLL is disabled, if you don't like it, but I thought that would keep the GUI a bit simpler in most use cases where the PLL isn't being used.

### DIFF
--- a/plugins/channelrx/chanalyzer/chanalyzer.cpp
+++ b/plugins/channelrx/chanalyzer/chanalyzer.cpp
@@ -139,7 +139,11 @@ void ChannelAnalyzer::applySettings(const ChannelAnalyzerSettings& settings, boo
             << " m_ssb: " << settings.m_ssb
             << " m_pll: " << settings.m_pll
             << " m_fll: " << settings.m_fll
+            << " m_costasLoop: " << settings.m_costasLoop
             << " m_pllPskOrder: " << settings.m_pllPskOrder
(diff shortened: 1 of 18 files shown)


f4exb:
If I understand right this is an new extra kind of PLL not a replacement of the existing one and thus an extension of the capabilities. Am I correct?

srcejon:
Yep - original PLL and FLL remain. You can choose between one of the 3 via a drop-down.

 The damping factor and loop gain controls I've added are only for the original PLL. I found some slight tweaking of these values can improve the lock. You might want to review the range of the controls.

srcejon:
Here's a pic

SigAnPLL

The new PLL row of controls is only visible when the padlock icon is pressed.

f4exb:
There is an error in Qt 5.11 (Docker image):
```
'PlaceholderText' is not a member of 'QPalette'
```
I vaguely remember this has been encountered before.

Issue #123, opened 2018-01-10
DATV: implement Digital Amateur TV modulator plugin


f4exb:
The title has it all... This is a feature issue for modulator only. The demodulator is followed up by this one #18

f4exb:
As suggested the TS could come via UDP (7 TS per frame) or file.

f4exb:
Implemented in v6.8.0

PR #810, opened 2021-03-18
DATV DVB-S modulator

srcejon:
Hi Edouard,

Here's a DATV modulator as per #123  TS can come from a file or via UDP.

Currently I've just implemented DVB-S, with BPSK and QPSK and selectable FEC code rates.

It now basically supports SDRangel to SDRangel broadcasts, but there are a few things that aren't quite right.

- BPSK demodulation only seems to work if Viterbi is checked.
- Audio doesn't seem to be properly synced on Windows or Linux (I don't see the demod buffer being used)
- When changing transport streams, it seems video for the second TS isn't played (unless demod channel is closed and reopened).

As of yet, I haven't yet determined whether these are issues in the modulator, demodulator or even transport stream (maybe pid related for the last one). I only started looking at DATV/DVB last week and haven't got another receiver to try out to compare, so not sure what to expect.


srcejon:
To get limesuite to build, I hacked:

#if !(defined(max)) && _MSC_VER
	// VC fix
//	#define max __max
#define max(a,b)            (((a) > (b)) ? (a) : (b))
#endif

in limesuite/src/lms7002m_mcu/MCU_File.h

f4exb:
I made a quick test and noticed the audio issue. It is still an excellent base. When playing the same TS file through a GNU Radio graph for DVB-S the audio from the DATV demod is not chopped.

For what concerns LimeSuite and MSVC I think the issue started when Appveyor upgraded to the latest version of MSVC 2019 because nothing has changed in the LimeSuite version being pulled neither in related cmake files in SDRangel. This is again the old story of "max" definition in MSVC. In the global compilation parameters the NOMINMAX definition is added (in main CMakeLists.txt) but since the last version of MSVC it seems disabled when compiling LimeSuite.

srcejon:
Just tried with GNU radio and I see a similar problem.

It might be related due to ffpmeg not quite generating a constant bitrate at the DVB datarate. (This modulator will try to insert null frames to match the rates if the TS bitrate is lower than the datarate - that's another possible source of a problem).

I thought the lack of buffering I was seeing in the demod might also be an issue. However, I hacked that to fill up to 50% before allowing it to be read, thinking this might help if there were some underruns, but it seems that buffer gets drained immediately (far faster than the TS bitrate). Will keep looking.

Yep, limesuite only became a problem for me after I upgraded VS.

srcejon:
The main problem appears to be because I still had the audio sample rate set as 44100 from looking at the other issue. Changing to 48000 is vastly improved, but not perfect.

I then tried viewing with ffplay via UDP rather than the integrated player. The audio there doesn't seem to have any problems. Also, the video and audio sync is better. With the integrated player, audio appears to be slightly delayed.

f4exb:
Issue https://github.com/myriadrf/LimeSuite/issues/344 opened for LimeSuite MSVC issue.

f4exb:
For the audio stuff, interestingly I am currently testing the modified Pluto FW named "perseverance" that turns the Pluto into a DVB-S2 transmitter. For the Rx part I am using SDRangel with the DATV plugin. I am transmitting a beacon with a continuous 440 Hz tone for the audio. The audio appears to be chopped (with the internal player). If I try to send a video from my smartphone using  Larix Broadcaster the audio lag is massive (about a minute).

f4exb:
With this commit https://github.com/f4exb/sdrangel/commit/debc5c74f18013b9553e0e1ad9bba0e7cac35c76 the audio latency is improved. There are still drops in some cases but I think globally the audio is improved.

f4exb:
Evariste, F5OEO has pointed me towards his library that contains code to modulate in DVB-S2: https://github.com/F5OEO/libdvbmod He said it contained ARM (Neon) intrisics but I haven't been able to spot just having a quick look so maybe it compiles on x86_64 as-is.

srcejon:
I had already tried integrating the DatvExpress S2 modulator that is based on - but the demod wouldn't lock (could be a problem of my own making, of course). I'll take a look, thanks.

srcejon:
Yeah - no different. I can commit it, if you want to try it.

f4exb:
sure, I just committed my latest changes on the demod to master. Should better manage the channelizer.

srcejon:
Ok, found the bug. Working now.

PR #826, opened 2021-03-31
Add DVB-S2 modulator

srcejon:
Adds G4GUO's DVB-S2 modulator.


f4exb:
I just did a small fix in cb657c72a to fix the file source looping. I think this is independent of DVB-S or DVB-S2 choice.

PR #839, opened 2021-04-12
SDRplay API support

srcejon:
This PR adds a sample source plugin that uses the SDRplay API V3, which should allow RSP1, RSP1A, RSP2, RSPduo and RSPdx support on Linux, Windows and presumably Mac.

Additionally, it supports additional RSP features not available in the existing driver, such as tuner and antenna selection, bias tee control, AM/MW/FM/DAB notch filters, AGC and external clock control.

Users will need to have installed the SDRplay driver and have the service running, but this should be able to be built without having it installed, as I've added the Windows headers and libs to sdrangel-windows-libraries (see PR https://github.com/f4exb/sdrangel-windows-libraries/pull/8 which will need to be merged for the Windows build to work) and the Linux headers as an external project (https://github.com/srcejon/sdrplayapi), in case the driver isn't already installed.

I've tested it with a RSPduo, but don't have access to any other devices.


srcejon:
Hmm. It builds OK for me when I use ENABLE_EXTERNAL_LIBRARIES=ON locally.

I see there are some extra build scripts called in .appveyor.yml (cmake/ci/build_cm256cc.sh etc) that do something similar to what I added in external/CMakeLists.txt. I can make something similar, but not sure why a different script is needed.

srcejon:
Can't why that is failing. Are you able to see the file system? Is the path to external/sdrplayapi/src/sdrplayapi/x86_64/libsdrplay_api.so.3.07 different?

f4exb:
`Submodule path 'external/windows': checked out 'a47e94ace126e9562fdc58d27fce95412e10d932'`
It looks like it takes the one before the merge.  I don; t know why. Maybe we should wait a little... ?

f4exb:
wait, wait wait... of course I have to update the commit in sdrangel...

f4exb:
MSVC succeeds!
Ubuntu fails but it seems related to an issue in Appveyor. This happened before.

f4exb:
Ubuntu build in Appveyoir is fixed by https://github.com/f4exb/sdrangel/commit/758c8a1bb0592176086f91f2be4f513b6840660d 

In fact for Linux package (Ubuntu) builds the "external libraries" builds are for libraries built from source using the `ExternalProject_Add` feature of cmake. SDRplay v3 is not in source code form and in this case the package has to be installed in the building image just like a package that would be installed via the distribution (`sudo apt install...`). This can be done with the script `cmake/ci/build_sdrplay.sh` ("build" is not the proper term in this case but this doesn't matter) that does automatically what the user would do manually. 

I checked the package produced by the CI chain and it does contain the new plugin for SDRplay v3. Of course people have to install SDRplay v3 in their system independently and this is expected.

srcejon:
A bit strange, as the external project was working for me locally (without having run the installer). However, that was linking with a specific version of the .so, so running the installer is better. (I didn't initially realise the installer would be able to be run on appveyor as it was installing services etc.). This seems like the best method - thanks!

f4exb:
What I did to compile from source was simply to install SDRplay v3 by cloning your Github repo: https://github.com/srcejon/sdrplayapi and run the install script there. Then touch the main `CMakeLists.txt` to trigger a re-run of the cmake thing and then `make`. `cmake` will pick up the system wide installed libraries and that's it. This is what is reproduced in the Appveyor's VM. The service is not needed at compile time just the libraries so it does not really matter if the service runs in the Appveyor's VM or not.

Issue #169, opened 2018-05-07
DAB/DAB+ plugin would be nice


beta-tester:
i have a feature request.

it would be nice to have a DAB/DAB+ plugin available for sdrangle.
see [welle.io](https://github.com/AlbrechtL/welle.io)

f4exb:
You are welcome to contribute.

texasyojimbo:
This kind of dovetails with my desire to have a "how to" guide for building/writing new modules.

I never got around to setting up on Linux, but I will try to get that done at some point. You had suggested in an old issue that we start with the SSB module as a template.

f4exb:
I will have a look at welle.io though as suggested. I always have found more difficult to adapt existing software than starting from scratch. The interesting point with welle.io is that it is Qt based.

mpbraendli:
welle-io diverged a bit from SDR-J: I've split GUI and backend a bit better, there's a way to instantiate the backend without a GUI and without Qt, and it is able to decode audio from all the programmes in the multiplex at the same time.

All this lives in the `next` branch for now. Let me know if you use it!

texasyojimbo:
In addition to welle, there is also this, referenced in context of DAB transmission on the rtl-sdr.com blog today:

https://www.rtl-sdr.com/creating-a-dab-radio-station-with-a-limesdr/
https://github.com/Opendigitalradio

The rtl-sdr.com blog article mentions using welle for receiving, but the ODR people also have a Linux tool using GTK+ called dablin for reception.

f4exb:
Deployed in v6.10.0

PR #850, opened 2021-04-16
Add DAB/DAB+ demodulator

srcejon:
Hi,

This PR adds a DAB/DAB+ demodulator to address #169. 

It should work on both Windows and Linux, however, it seems there is a base SDRangel issue on Windows which can cause it not to work in some instances (SampleSinkFifo::write: overflow when there are no demodulators). I'll file a bug for that later.

Fairly basic for now.


srcejon:
To build on windows, an update to sdrangel-windows-libraries is needed: https://github.com/f4exb/sdrangel-windows-libraries/pull/9

f4exb:
It looks like the Ubuntu build in Appveyor is failing because it creates `libdab_lib.so` while it expects `libdab.so` later:
```
[00:08:02] [39/39] Linking CXX shared library /home/appveyor/projects/sdrangel/obj-x86_64-linux-gnu/external/lib/libdab_lib.so
```
then
```
[00:23:19] /usr/bin/ld: cannot find -ldab
```
I don't know why this `_lib` suffix is appended. I suppose this should be for MSVC.

srcejon:
Thanks. It's odd, but that's just what the DAB library author called it, which was written for Linux (My fork to make it compile with MSVC didn't change it).  I just used the wrong name in external/CMakeLists.txt. Hopefully fixed now.

f4exb:
Never mind the odd name. It works!

Issue #858, opened 2021-04-20
AIS Demodulator


choupette-change:
Evolution of a new demodulator: Have you planned to make an ais demodulator (vhf position of ships), like the adsb demodulator, soon, with the card already present in sdrangel ? I am already using "aisdecoder" but everything integrated into sdrangel that would be very good...

choupette-change:
I test the AIS demodulator + feature on Ubuntu 20 10 (rtlsdr) it's ok for me (I live near the west coast of France) . Where did you find the icons? Good work . Can you predict the units (km, s, ..)?

srcejon:
If you can find/make some better and more diverse icons, I will update them. I couldn't find many that were top down, which is needed to draw them according to their heading.

f4exb:
Works great!
![image](https://user-images.githubusercontent.com/6192319/117778823-53e27a80-b23e-11eb-9627-c5404ac50b20.png)

srcejon:
The names don't look right though. The @ character is a 0.  I strip them from the start and end, but hadn't seen any in the middle, which is a little suspect (although the callsign etc seem ok). If you can make an sdriq file, I can take a look. Thanks.

srcejon:
Actually, they all look like small boats rather than ships, which means the problem is with Class B messages, and I think I see the problem.

srcejon:
Thanks - looks like there's another problem with Static Data Reports. Not such a variety of message type around here.

choupette-change:
ship names seem correct to me
![image](https://user-images.githubusercontent.com/68333247/117794954-02da8280-b24e-11eb-815e-befbbe8e68e5.png)
![image](https://user-images.githubusercontent.com/68333247/117795369-695fa080-b24e-11eb-8816-ef7129ce4940.png)

Valstee:
So Nice, Thanks Srcejon !

f4exb:
Implemented

PR #891, opened 2021-05-07
Add AIS mod, demod and feature

srcejon:
As per #858.

I don't live near the sea, so only done a quick bit of testing by the Thames. Could do with some better icons.


f4exb:
I will make it a pleasure to test it on the French Riviera! :sunglasses:

choupette-change:
I test the AIS demodulator + feature on Ubuntu 20 10 it's ok for me (I live near the west coast of France) . Where did you find the icons? Good work . Can you predict the units (km, s, ..)?

PR #906, opened 2021-05-20
Antenna calculators

srcejon:
This Antenna Tools feature adds a couple of antenna calculators: 

half wave dipole calculator - calculating length from frequency and vice-versa
parabolic dish calculator - calculate focal length, beamwidth and gain.

I suspect you might think this is a bit bloaty... but I'm always googling the formulas, so thought it would be easier to to calculate automatically based on the device set's frequency. There's a few others I can add if you don't think it's too superfluous.


f4exb:
It is a bit surprising in a SDR software but feature plugins are there also for features that may not seem essential to SDR business. So why not...

Issue #285, opened 2019-02-03
Open .wav files recorded from SDRSharp


loic2665:
Hi,

So I just discovered SDRAngel, that is a beautiful and complete software (Big thanks for DSD decoder built in !)

But I recorded a Raw signal from SDRSharp and it output a `.wav` file.
I tried to open it via SDRAngel (from fileSource option), but no luck i needed a `.sdriq` file, and not a `.wav`

I added a `.sdriq` extention to the file but obviously, this didn't worked. I read about a File input plugin,  from the issue #152  but I have no idea where to enable it (F7 to record !? F8 to stop ?? (still a `.sdriq` file... and no reading option...) 

Could you please guide me a bit ? Or is there a way to convert a `.wav` file to a `.sdriq` file ? Is there metadata to the `.sdriq` file ?

Thanks.

f4exb:
Hi,

yes there is metadata in the .sdriq files. You will find a utility to re-create this meta data here: https://github.com/f4exb/sdrangel/tree/master/rescuesdriq

I suppose the .wav files in question are actually I/Q samples saved as audio stereo. You will need to first convert it to raw format (2 channel S16LE) using a utility like `sox`. You may loose some dynamic due to the conversion to 16 bit samples. .sdriq files may also be saved as 24 bit over 32 bit samples (8 zero bit MSB) but I am not sure `sox` can convert to this format.

Of course you must know at least the original sample rate. The other parameters like center frequency and initial timestamp are not critical to operation.

loic2665:
Hi, thanks for your time, and sorry for my late answer.

Yes, this is I/Q sample. I'm less sure about the stereo feature.
But yes, I know the sample rate, we can see it through Audacity too. (For stereo channel too)

This seems a bit tricky to open a simple `.wav` file. Maybe to improve?

But I will take a look. And I will give you a feedback.

Thanks.

f4exb:
.wav files are not supported. You need the metadata header.

PR #907, opened 2021-05-21
Add .wav file support for File Input and File Sink

srcejon:
This PR adds signed 16-bit PCM  .wav file support to the File Input and File Sink plugins.

I find this useful, as it's easier to use these files with other tools for analysis and editing.

While .wav can support 32-bit, I made this 16-bit as an option to have a reduced file size. Capturing data at several MSa/s results in very large files and at these sort of rates I don't think the extra bits are really useful. (In the future, could add an option to select 8/16/32-bit capture for either file type, but that needs a little more work).

This supports the .wav auxi header used by some other signal processing s/w to save centre frequency and start time, so functionality shouldn't be reduced in comparison to .sdriq. It can also work this info out from the filename used by a couple of other SDR programs, that don't use the auxi header.

Implements #285

### DIFF
--- a/plugins/channelrx/filesink/filesinkgui.cpp
+++ b/plugins/channelrx/filesink/filesinkgui.cpp
@@ -500,7 +500,7 @@ void FileSinkGUI::on_showFileDialog_clicked(bool checked)
         this,
         tr("Save record file"),
         m_settings.m_fileRecordName,
-        tr("SDR I/Q Files (*.sdriq)")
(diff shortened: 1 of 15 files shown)


f4exb:
Good idea! Do you plan the other way round (File Source and File Output)? Or I can do it myself later based on your work that contains the essential bits. This is probably a less common usage but would be just for completion.

srcejon:
Ah, didn't think of those. File Source should be trivial, but File Output probably needs a bit more re-work.

PR #908, opened 2021-05-21
Add .wav file support to File Source plugin

srcejon:


f4exb:
Excellen! In fact this is the most useful in order to use an I/Q .qav file as a source for transmission. File Output usage would be to create a .wav fle from modulator(s) and would be useful only to produce test files.

PR #913, opened 2021-05-30
Use shared_ptr to keep track of DeviceScan across multiple collections

### LINKED ISSUE #681: SDRangel with 2 PlutoSDR
Please help! Cannot start 2 plutos in SDRangel simultaneously. One Pluto starts, but the second doesn't.
No spectrum/waterfall, showing strangeactual gain values, no rssi, and no temp. The video below shows better than 1000 words what I mean.
What I'm doing wrong?
The version 4.11.9 works like expectedm but the last version from release page has these strange issues.
Both plutos are v0.32 firmware, but with v0.31 is the same situation.
https://www.youtube.com/watch?v=Lh0w0G70_sA

ChiefGokhlayeh:
Fixes #681.

Relying on memory allocated by `std::vector` via direct pointers is dangerous, as vectors may move buffers during resize. This fix puts the `DeviceScan` allocation into a `str::shared_ptr`, which 1) allocates the memory on heap and 2) keeps track of the memory through reference counting. Thus it is safe to store the `shared_ptr` across different vectors/maps.

### DIFF
--- a/devices/plutosdr/deviceplutosdrscan.cpp
+++ b/devices/plutosdr/deviceplutosdrscan.cpp
@@ -19,6 +19,7 @@
 #include <cstdio>
 #include <cstring>
 #include <regex>
+#include <memory>
 #include <iio.h>
 
 #include <QtGlobal>
@@ -49,6 +50,7 @@ void DevicePlutoSDRScan::scan()
     }
 
     m_scans.clear();
+    m_scans.reserve(num_contexts);
 
     for (i = 0; i < num_contexts; i++)
     {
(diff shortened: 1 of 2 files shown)


ChiefGokhlayeh:
Looking back on my code, a slimmer implementation using raw pointers, i.e. ...
```c++
std::vector<DeviceScan*> m_scans;
```
... may also be feasible. However, it requires that we manually `delete` all pointers stored in `m_scans` upon destruction.

I could easily change the implementation. I'll let you decide! :)

PR #916, opened 2021-06-04
Delete ADSBDemod::m_worker after removing sink from DSP

### LINKED ISSUE #915: Segmentation Fault when closing ADS-B panel while Sampling Device is still running
I noticed a segmentation fault whenever closing the ADS-B Demodulator panel while having the sampling device still running.

Steps to reproduce:

1) Open sdrangel
2) Select "Sampling Device" `TestSource`
3) Add ADS-B Demodulator to Channels
4) Start `TestSource`
5) Close ADS-B Demodulator panel without stopping `TestSource`

If `TestSource` is stopped before closing ADS-B Demodulator everything continues to work fine.

I traced the issue down to an ~"use-after-delete"~ "use-after-_half_-delete" by `DSPDeviceSourceEngine`:
https://github.com/f4exb/sdrangel/blob/3e9b4a4dee67dd37a9bcdd90c00c5141550c189d/sdrbase/dsp/dspdevicesourceengine.cpp#L595-L597

In this case `sink` is the already _half_ deleted [`ADSBDemod`](https://github.com/f4exb/sdrangel/blob/3e9b4a4dee67dd37a9bcdd90c00c5141550c189d/plugins/channelrx/demodadsb/adsbdemod.cpp#L124-L130) object. The `delete` is called before that by `DeviceUISet::handleChannelGUIClosing()` on UI-thread:
https://github.com/f4exb/sdrangel/blob/de23efe635043867b914faa2977228e0780d4121/sdrgui/device/deviceuiset.cpp#L439

-- Tested on [latest master](https://github.com/f4exb/sdrangel/commit/3e9b4a4dee67dd37a9bcdd90c00c5141550c189d) following v6.13.0.

ChiefGokhlayeh:
Fixes #915.

As described in the issue, during deconstruction of [`ADSBDemod`](https://github.com/f4exb/sdrangel/blob/3e9b4a4dee67dd37a9bcdd90c00c5141550c189d/plugins/channelrx/demodadsb/adsbdemod.cpp#L81) a segmentation fault might occur, if `DSPDeviceSourceEngine` calls [`sink->stop()`](https://github.com/f4exb/sdrangel/blob/3e9b4a4dee67dd37a9bcdd90c00c5141550c189d/sdrbase/dsp/dspdevicesourceengine.cpp#L596) _after_ `m_worker` was deleted.

This PR fixes the issue by deleting `m_worker` only after [`m_deviceAPI->removeChannelSink(this)`](https://github.com/f4exb/sdrangel/blob/3e9b4a4dee67dd37a9bcdd90c00c5141550c189d/plugins/channelrx/demodadsb/adsbdemod.cpp#L90) is called. This call would otherwise invoke `sink->stop()` with a broken `m_worker` somewhere down the line.

### DIFF
--- a/plugins/channelrx/demodadsb/adsbdemod.cpp
+++ b/plugins/channelrx/demodadsb/adsbdemod.cpp
@@ -83,11 +83,11 @@ ADSBDemod::~ADSBDemod()
     if (m_worker->isRunning()) {
         stop();
     }
-    delete m_worker;
     disconnect(m_networkManager, SIGNAL(finished(QNetworkReply*)), this, SLOT(networkManagerFinished(QNetworkReply*)));
     delete m_networkManager;
     m_deviceAPI->removeChannelSinkAPI(this);
     m_deviceAPI->removeChannelSink(this);
+    delete m_worker;
     delete m_basebandSink;
     delete m_thread;
 }


ChiefGokhlayeh:
I scanned through the other `channelrx` plugins and didn't find any bugs of the same type.

f4exb:
It does not hurt and incidentally solves the issue. Moreover now the deletion occurs in reverse order of the news (`m_networkManager` is created last).  But as I said in my comments of #915 in the first place the various workers should take care of their running state before doing the stopping job (same thing for start) because their `stopWork` method may be called several times legitimately.

Issue #953, opened 2021-07-06
Web API update for "Satellite Radio Control" tabs in "Satellite Tracker" feature


Valstee:
Hi @f4exb, @srcejon, 
Is it possible to PATCH "Satellite Radio Control" settings (in the Satellite tracker feature) using the web API ?
It will be perfect to control my small "raspberry ground station" with sdrangel server !
Thanks a lot for your answer,
Valentin

srcejon:
Not at the moment.  The radio control settings aren't yet accessible.

Valstee:
Ok ! Thanks for your answer @srcejon, it could be great  ;-)

srcejon:
This should now work. You can add settings with something like:

    curl -X PATCH "http://127.0.0.1:8091/sdrangel/featureset/0/feature/0/settings" -d '{"featureType": "SatelliteTracker",  "SatelliteTrackerSettings": {  "deviceSettings": [ {  "satellite": "ISS",  "deviceSettings": [ { "deviceSet": "R0", "doppler": [0], "frequency": 0, "presetDescription": Sat", "presetFrequency": 145.825, "presetGroup": "ISS Digi", "startOnAOS": 1, "startStopFileSinks": 1, "stopOnLOS": 1}] } ] }}'

Valstee:
Yes really good! I confirm you that it works! A huge thanks! 🥇 
Can I close the issue or are there still things to do? (because you said "start to add")

srcejon:
Yep, you can close - another patch was needed - but I forgot to tag it with this issue number.

PR #958, opened 2021-07-08
Start to add support for device settings in Satellite Tracker web API

srcejon:
This patch starts to add support for device settings in Satellite Tracker web API for #953

However, there is a problem, in that it gets a nullptr exception, possibly because one of the Swagger nested objects has not had init() called (at least, it looks similar to the previous problem). 

I can't see how a similar workaround in webapirequestmapper.cpp could work (i.e.)

            featureSettings->getSatelliteTrackerSettings()->init();

As this is for a nested object within a list within the satellite tracker settings, that is only created when the JSON is parsed.

However, I've noticed that the SWGModelFactory.h that you've commited [here](https://github.com/srcejon/sdrangel/commit/70167c0ed81adb98e3aa7c4d4b280ca020798180#diff-cb8fb6955ea31ee4855a9e4c6897585a9f2e5912a544e57d66611cd82c9bef61) is different to mine. It seems to include calls to init().

E.g.:

     if(QString("SWGADSBDemodReport").compare(type) == 0) {
          +SWGADSBDemodReport *obj = new SWGADSBDemodReport();
          +obj->init();
          +return obj;
          -return new SWGADSBDemodReport();
        }

What do I need to do to get the swagger code generated like this? I believe my repo is up to date.


f4exb:
Yes, I have updated the Swagger code generator: https://github.com/f4exb/swagger-codegen/tree/sdrangel
You simply need to rebuild a Docker image here: https://github.com/f4exb/sdrangel/tree/master/swagger/docker/codegen with the `build.sh` script as the `Dockerfile` pulls the head of `sdrangel` branch of the codegen repo.

srcejon:
Ah, thanks. That's better. 

Looks like I need to make an additional change to handle missing properties.

PR #960, opened 2021-07-09
Add Pager demodulator plugin

srcejon:
This PR adds a simple POCSAG Pager demodulator plugin. Supports 512, 1200 and 2400 baud.


nir4x6fk:
Hello,
Is it possible to change the language in POCSAG messages?

srcejon:
https://github.com/f4exb/sdrangel/pull/966 added support for being able to change the character set (press the keyboard icon).

nir4x6fk:
Thank you for reply. sorry... But I can't see the keyboard icon on the POCSAG decoder  (after adding channel pager demodulator)

srcejon:
That change isn't in the most recent release yet.

You'll either need to build it yourself from source, or pick up the test build here: https://ci.appveyor.com/api/buildjobs/9ci2x6gxpt4wj379/artifacts/build%2Fsdrangel-6.16.1-2-g990c279-win64.exe

nir4x6fk:
Thank you. GREAT!

srcejon:
If it works for you, let me know what mapping you use, then I can possibly add it as a preset.

nir4x6fk:
I use hebrew left to right. (it is strange, as hebrew reading is right to left) other whey it appear in opossite direction. most of the messages looks like encrypted as it types "gibrish"... But the messages that are not (as I'm thing that are not encrypted), types OK in hebrew, and are readable.

srcejon:
Ok, well it's good some are readable.

The left-to-right / right-to-left option just flips the order of the received characters. If you need to use left-to-right, then that suggests the characters are transmitted last first - so this option isn't actually needed. I was assuming they would be transmitted right-most character first, and thus needed reversing (I do not have a document that says how Hebrew should work).

Within the POCSAG specification, there isn't support for encryption. If you are receiving garbage messages, this could be for a couple of reasons:

- The decoder is decoding noise as a message
- You need to try a different value for the Decode option, which determines whether messages are numeric or alphanumeric
- The signal isn't quite strong enough and some bits have been received incorrectly
- There's a bug in the decoder

If there are any parity errors indicated (in the Even PE or BCH PE) columns, then it's probably a case of some bits not being demodulated correctly.

Feel free to upload some screenshots or IQ files showing the problem.

nir4x6fk:
I started to play with the POCSAG keyboard icon. 
1. In hebrew are 22 letters + 5 "final letters". I tried to add the  letters as Unicode and 7bit, but after closing the keyboard icon and open it again, all what I added disappear. 
2. It is not possible add in "user" characters, and to save.

nir4x6fk:
image

srcejon:
It works for me.

Select preset Hebrew 
Change reading order to left to right
Press +
Enter 76 and and 5e6
Press +
Enter 77 and and 5e7
...
Press OK
Save the preset
Press the keyboard Icon again and the preset should be User with your extra characters at the bottom

image

nir4x6fk:
Hello,
Yes this what I did.
It works also for me :). after I add the new characters, I pressed the keyboard icon. I did not see any "save" button, but naw the characters are saved.

nir4x6fk:
Well after close SDRANGEL,.   today I start it again, and it returned to default. I need to enter the characters again...
I also did not see any "save" option...
Sorry... :(

srcejon:
The settings should be saved when you create a preset containing the pager demod (along with the demod's other settings).

See section 5 presets here: https://github.com/f4exb/sdrangel/blob/master/sdrgui/readme.md

I can probably add those 5 extra characters as standard to the Hebrew character set - if you think that makes sense.

nir4x6fk:
Thanks' I will try that.
To add the those 5 extra characters as standard to the Hebrew character set, it makes sense, as they are part of the hebrew alphabet.
Thank you

Issue #980, opened 2021-08-04
Higher resolution timestamp in .sdriq format


ChiefGokhlayeh:
I was wondering if we could increase the resolution of the `FileRecord` timestamp to milliseconds.
https://github.com/f4exb/sdrangel/blob/14e75232df072a250ac5e8888e804fdb37439f80/sdrbase/dsp/filerecord.h#L40
Currently it's defined as a 64-bit integer which stores a UNIX timestamp in seconds resolution. Is there some specific reasoning behind that?

The change in code is fairly trivial, but I know changing a file-format once it's established _is not_. I'd still like to encourage a discussion about it.

Context:
In my use-case I want to synchronize two recordings from two separate coherent receivers. The signal I'm capturing (LTE-based 5G Broadcast) provides a synchronization sequence in 40ms intervals, which I'm able to decode. But the current timestamp resolution makes synchronizing across 1sec-boundaries unnecessarily hard.

ChiefGokhlayeh:
Feature implemented and merged.

PR #981, opened 2021-08-04
Higher resolution timestamp in sdriq format

ChiefGokhlayeh:
As mentioned in #980 here's my take on increasing timestamp resolution in file-headers.

While going through the references I found some golang code for converting old sdriq files to the newer format. I tried changing that as well, but I'm not too familiar with go, so please someone vet my code.


f4exb (approved):

ChiefGokhlayeh:
Did some manual testing with the filesink and fileinput plugins. Assuming I didn't miss anything I'll mark this PR ready-for-review.

Don't feel pushed to merge this PR. Feel free to decline, if you think file format compatibility should be kept.

PR #983, opened 2021-08-12
Fix wrong start-datetime display in GUI for old file records

ChiefGokhlayeh:
Looking to improve backward compatibility with the previous `FileRecord::Header` format I provided a `FileRecord::timeStampToDateTime()` function, which takes in a `quint64` and tries to convert it into a sensible `QDateTime`. It uses a simple (and kinda hack'ish) heuristic to decide whether the timestamp is in seconds or milliseconds (old vs. new format).

This should now fix date-time displays in GUI when opening old file records.

Subsequent changes include switching the use of `quint64` timestamps as time-keepers to `QDateTime`. Done so, as to ensure clean interfaces without ambiguities (no more "seconds, milliseconds, which is it?" guessing).
Applied in:

* channeltx/filesource
* samplesink/fileoutput
* samplesource/fileinput
* samplesource/localinput
* samplesource/remoteinput
* samplesource/sigmffileinput

Most of which was done automated through an IDE.

### DIFF
--- a/plugins/channeltx/filesource/filesource.cpp
+++ b/plugins/channeltx/filesource/filesource.cpp
@@ -471,7 +471,7 @@ void FileSource::webapiFormatChannelReport(SWGSDRangel::SWGChannelReport& respon
     qint64 t_msec = 0;
     quint64 samplesCount = m_basebandSource->getSamplesCount();
     uint32_t fileSampleRate = m_basebandSource->getFileSampleRate();
-    quint64 startingTimeStamp = m_basebandSource->getStartingTimeStamp();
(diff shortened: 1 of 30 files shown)


ChiefGokhlayeh:
If the delta is too big or the hack too hack'ish, I'd happily adapt any change-requests, obviously.

ChiefGokhlayeh:
|-Before:-|-After-|
|-|-|
|old | new |

ChiefGokhlayeh:
It's probably better to convert file records offline, rather than relying on heuristics inside the application.

PR #1026, opened 2021-10-30
Fix #897 - Add support for vertically expanding widgets to RollupWidget

### LINKED ISSUE #897: Rollup widget doesn't appear to support vertical expanding policy.
It's seems like the rollup widget doesn't support the vertical expanding policy. I think it would be useful if it did, so that various features and (perhaps some specific channels), would expand to fill the available space. E.g:

image

The table and map should really fill the available space so that max amount of data is visible. Horizontal expansion seems to work. It could be I haven't set some property correctly, of course. But I think I have expanding set in all of the AIS widgets.

srcejon:
And adjust some features / channels to take advantage of it.

### DIFF
--- a/plugins/channelrx/chanalyzer/chanalyzergui.ui
+++ b/plugins/channelrx/chanalyzer/chanalyzergui.ui
@@ -10,6 +10,12 @@
     <height>778</height>
    </rect>
   </property>
+  <property name="sizePolicy">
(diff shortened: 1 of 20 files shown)


f4exb:
This is ugly
image

f4exb:
Sorry, I am going to revert this

PR #1108, opened 2022-01-19
APRS Mic-E Message Decoding

ooglek:
I kept seeing these messages in APRS and didn't understand them. So I read the [APRS Mic-E Spec](http://www.aprs.org/doc/APRS101.PDF) and attempted to implement it.

This is my first attempt, both at implementing Mic-E and writing code in C++. This code isn't tested. I honestly don't know how to test it yet. I'm hoping to get some advice, feedback, and guidance on how best to make this work! Happy to hop on a call or video chat or message on how best to make this happen.

### DIFF
--- a/sdrbase/util/aprs.cpp
+++ b/sdrbase/util/aprs.cpp
@@ -1012,7 +1012,7 @@ bool APRSPacket::parseMicE(QString& info, int& idx, QString& dest)
     QString latDigits = '';
     QString messageBits = '';
     int messageType = 0; // 0 = Standard, 1 = Custom
-    int latitudeOffset = 0;
+    int longitudeOffset = 0;
     float latitudeDirection = -1; // Assume South because North is easier to code
     float longitudeDirection = 1;
 
@@ -1023,7 +1023,7 @@ bool APRSPacket::parseMicE(QString& info, int& idx, QString& dest)
             if (inRange(48, 57, charInt)) {
                 latDigits.append(static_cast<char>(dest[i] % 48));
             } else if (inRange(65, 74, charInt)) {
-                latDigits.append(static_cast<char>(dest[i] % 48));
+                latDigits.append(static_cast<char>(dest[i] % 65));
             } else if (inRange(80, 89, charInt)) {
                 latDigits.append(static_cast<char>(dest[i] % 80));
             } else {
@@ -1052,7 +1052,7 @@ bool APRSPacket::parseMicE(QString& info, int& idx, QString& dest)
             // Longitude Offset
             if (i == 4) {
                 if (inRange(65, 75, charInt))
-                    latitudeOffset = 100;
+                    longitudeOffset = 100;
             }
 
             // Longitude Direction
@@ -1063,6 +1063,7 @@ bool APRSPacket::parseMicE(QString& info, int& idx, QString& dest)
         }
 
         m_latitude = (((float)latDigits.mid(0, 2)) + latDigits.mid(2, 2)/60.00 + latDigits.mid(4, 2)/60.0/100.0) * latitudeDirection;
+        m_hasPosition = true;
 
     } else
         return false;
@@ -1086,11 +1087,9 @@ bool APRSPacket::parseMicE(QString& info, int& idx, QString& dest)
         && (charToInt(info, idx+7) == 47 || charToInt(info, idx+7) == 92)
        )
     {
-        // Longitude
-        int deg = charToInt(info, idx) - 28;
+        // Longitude, plus offset encoded in the AX.25 Destination 
+        int deg = (charToInt(info, idx) - 28) + longitudeOffset;
         // Destination Byte 5, ASCII P through Z is offset of +100
-        if (inRange(80, 90, charToInt(dest, 4)))
-            deg += 100;
         if (inRange(180, 189, deg))
             deg -= 80;
         if (inRange(190, 199, deg))
@@ -1118,8 +1117,9 @@ bool APRSPacket::parseMicE(QString& info, int& idx, QString& dest)
     } else
         return false;
 
-    return true;
     // Altitude
     // #TODO
+
+    return true;
 }
 
--- a/sdrbase/util/aprs.cpp
+++ b/sdrbase/util/aprs.cpp
@@ -143,6 +143,10 @@ bool APRSPacket::decode(AX25Packet packet)
                 parseTimeMDHM(packet.m_dataASCII, idx);
                 parseWeather(packet.m_dataASCII, idx, true);
                 break;
+            case '`': // Mic-E Information Field Data (current)
+            case '\'': // Mic-E Information Field Data (old)
+                parseMicE(packet.m_dataASCII, idx, m_to);
+                break;
             case '{': // User-defined APRS packet format
                 break;
             default:
@@ -166,6 +170,12 @@ int APRSPacket::charToInt(QString&s, int idx)
 {
     char c = s[idx].toLatin1();
     return c == ' ' ? 0 : c - '0';
+}
+
+// Mic-E Byte in Range
+bool APRSPacket::inRange(unsigned low, unsigned high, unsigned x)
+{
+ return (low <= x && x <= high);
 }
 
 bool APRSPacket::parseTime(QString& info, int& idx)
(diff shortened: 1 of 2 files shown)


sdrbase/util/aprs.cpp
```diff
@@ -988,3 +998,128 @@ bool APRSPacket::parseTelemetry(QString& info, int& idx)
...
+    // Mic-E Location data is encoded in the AX.25 Destination Address
+    if (dest.length() < 6)
+        return false;
+
+    // Mic-E Information data is 8 bytes minimum, 13-14 with altitude
+    if (info.length() < idx+8)
+        return false;
+
+    QString latDigits = '';
+    QString messageBits = '';
```
ooglek, line 1013: I wasn't sure if this should be a new `m_mice_messagebits` variable or if I should just shove it in comments somewhere. It's a lookup table with 7 options plus 1 emergency tag.

sdrbase/util/aprs.cpp
```diff
@@ -988,3 +998,128 @@ bool APRSPacket::parseTelemetry(QString& info, int& idx)
...
+    float longitudeDirection = 1;
+
+    QRegExp re("^[A-LP-Z0-9]{3}[L-Z0-9]{3}.?$"); // 6-7 bytes
+    if (re.exactMatch(dest)) {
+        for (int i = 0; i < 6; i++) {
+            int charInt = charToInt(dest, i);
+            if (inRange(48, 57, charInt)) {
+                latDigits.append(static_cast<char>(dest[i] % 48));
+            } else if (inRange(65, 74, charInt)) {
+                latDigits.append(static_cast<char>(dest[i] % 48));
```
ooglek, line 1026: oops this should be mod 65...

sdrbase/util/aprs.h
```diff
@@ -461,6 +461,7 @@ struct SDRBASE_API APRSPacket {
     int charToInt(QString &s, int idx);
     bool parseTime(QString& info, int& idx);
     bool parseTimeMDHM(QString& info, int& idx);
+    bool inRange(unsigned low, unsigned high, unsigned x);
```
f4exb, line 464: Obviously this is a very simple method just there to avoid writing the same condition over and over. In this case the body could be put here as an inline method.

sdrbase/util/aprs.cpp
```diff
@@ -168,6 +172,12 @@ int APRSPacket::charToInt(QString&s, int idx)
     return c == ' ' ? 0 : c - '0';
 }
 
+// Mic-E Byte in Range
+bool APRSPacket::inRange(unsigned low, unsigned high, unsigned x)
+{
+ return (low <= x && x <= high);
```
srcejon, line 178: Indentation should be 4 spaces, rather than 1.

sdrbase/util/aprs.cpp
```diff
@@ -988,3 +998,128 @@ bool APRSPacket::parseTelemetry(QString& info, int& idx)
...
+        speed += floor(decoded_speed_course); // Speed in 1 kt units, added to above
+        int course = (((charToInt(info, idx+4) - 28) % 10) * 100) % 400;
+        int course += charToInt(info, idx+5) - 28;
+
+        m_longitude = (((float)deg) + min/60.00 + hundreths/60.0/100.0) * longitudeDirection;
+        m_hasPosition = true;
+        m_symbolTable = info[idx+7].toLatin1();
+        m_symbolCode = info[idx+6].toLatin1();
+        m_hasSymbol = true;
+    } else
```
srcejon, line 1117: I'd recommending using {} around the else if you've used them around the if {}

sdrbase/util/aprs.cpp
srcejon: single quotes are for character literals. "" is for strings.

srcejon:
Thanks for the patch.

With regards to functional testing, I would suggest a few things:

- First, make sure the APRS decoder continues to decode non-Mic-E packets OK. (It should do - but always worth testing)
- Then, for testing your new Mic-E functionality, you should be able to see if it can successfully decode some of the packets that you initially saw that encouraged you to write this. You can record and replay packets using IQ files (Via the File Sink and File input plugins) in case it doesn't work first time.
- If you want to do even more testing, you could transmit your own Mic-E encoded packets using the Packet Modulator, to test that they are decoded properly, covering all the corner cases you can think of.

sdrbase/util/aprs.cpp
ooglek: Fixed

sdrbase/util/aprs.cpp
ooglek: Fixed

sdrbase/util/aprs.h
ooglek: Moved to inline at the top of the file. Oops need to remove this from the `.h`

ooglek:
Addressed the feedback, thanks!

I've read all the Developer-notes and building docs. Unfortunately I'm on a Mac and I'm not sure how to build SDRAngel on the Mac so I can test, despite it being installed and running on my Mac from MacPorts.

I'd love to learn how to build and test locally so I can feel more confident about the PR, just need a bit more guidance, such as which C++ compiler to use, any information on using Vim + ALE for development, how to compile and test on a Mac, frequent pitfalls, obscure compiler warnings to ignore or take seriously, etc. 

In the meantime, I'll grab the SDRangel Docker projects and try testing my code there.

srcejon:
You should be able to build on a Mac following the Linux build instructions: https://github.com/f4exb/sdrangel/wiki/Compile-from-source-in-Linux - aside from the apt-get package installs.

ooglek:
Wicked. Some feedback on the build process for my build on a Mac with sdrangel installed via MacPorts for future viewers:

1. Boost -- Had to add `-DBOOST_ROOT=/opt/local/libexec/boost/1.76` -- got error `Could NOT find Boost (missing: Boost_INCLUDE_DIR)`
2. SerialDV -- Had to change to `-DSERIALDV_DIR=/opt/local` -- got error `Could not find SerialDV`

And now finally I'm getting all the warnings and errors I had hoped in my code! I'll go fix those. :-)

ooglek:
OK, that took some work, but I think it is good! Please do a bit more of a sanity check. 

I've been decoding for a few hours on and off, I think I've found all the bugs on the Lat/Lon, Course, Speed, Altitude, and Status. There's some more work to do on handling some edge cases that add characters, as well as adding the Telemetry support, but despite a few hours of logging, I haven't seen a Telemetry report via Mic-E yet. 

This has been fun!

sdrbase/util/aprs.cpp
```diff
@@ -16,10 +16,25 @@
 ///////////////////////////////////////////////////////////////////////////////////
 
 #include <QRegExp>
+#include <QRegularExpression>
```
ooglek, line 19: QT5 Docs stated to use QRegularExpression over QRegExp

> Note: In Qt 5, the new QRegularExpression class provides a Perl compatible implementation of regular expressions and is recommended in place of QRegExp.

https://doc.qt.io/qt-5/qregexp.html

sdrbase/util/aprs.cpp
```diff
@@ -16,10 +16,25 @@
...
+#include "util/units.h"
+
+inline bool inRange(unsigned low, unsigned high, unsigned x)
+{
+    return (low <= x && x <= high);
+}
+
+// Required for Mic-E Decoding
+inline int charToIntAscii(QString&s, int idx)
+{
```
ooglek, line 33: The other one returned `0` for a space, which breaks Mic-E course and speed use the ASCII range 28 to 127, which includes space (dec 32) and thus I rewrote to be more faithful.

I'm happy to rewrite the other one and have it return correctly and fix the space vs 0 issue where needed... otherwise this works.

sdrbase/util/aprs.cpp
```diff
@@ -30,8 +45,9 @@ bool APRSPacket::decode(AX25Packet packet)
     if ((packet.m_type == "UI") && (packet.m_pid == "f0") && (packet.m_dataASCII.length() >= 1))
     {
         // Check destination address
-        QRegExp re("^(AIR.*|ALL.*|AP.*|BEACON|CQ.*|GPS.*|DF.*|DGPS.*|DRILL.*|DX.*|ID.*|JAVA.*|MAIL.*|MICE.*|QST.*|QTH.*|RTCM.*|SKY.*|SPACE.*|SPC.*|SYM.*|TEL.*|TEST.*|TLM.*|WX.*|ZIP.*)");
-        if (re.exactMatch(packet.m_to))
+        QRegularExpression re("^(AIR.*|ALL.*|AP.*|BEACON|CQ.*|GPS.*|DF.*|DGPS.*|DRILL.*|DX.*|ID.*|JAVA.*|MAIL.*|MICE.*|QST.*|QTH.*|RTCM.*|SKY.*|SPACE.*|SPC.*|SYM.*|TEL.*|TEST.*|TLM.*|WX.*|ZIP.*)");
+        QRegularExpression re_mice("^[A-LP-Z0-9]{3}[L-Z0-9]{3}.?$"); // Mic-E Encoded Destination, 6-7 bytes
+        if (re.match(packet.m_to).hasMatch() || re_mice.match(packet.m_to).hasMatch())
```
ooglek, line 50: The `m_to` field is used to encode the latitude, message type, cardinal direction of lat and lon, as well as the longitude offset bit. 

I considered making it separate, which I'm happy to, as this might match things it should not.

sdrbase/util/aprs.cpp
```diff
@@ -156,9 +176,14 @@ bool APRSPacket::decode(AX25Packet packet)
             }
 
             return true;
+        } else {
+            qDebug() << "APRSPacket::decode: AX.25 Destination did not match known regexp " << m_to;
```
ooglek, line 180: I added this as some of the other Features throw out debug messages when there are failures, such as 

> PacketDemodSink::processOneSample: CRC mismatch: 69e7 c228

sdrbase/util/aprs.cpp
```diff
@@ -988,3 +1013,196 @@ bool APRSPacket::parseTelemetry(QString& info, int& idx)
...
+
+    QHash<QString, QString> messageTypeLookup;
+    messageTypeLookup["111"] = "Off Duty";
+    messageTypeLookup["110"] = "En Route";
+    messageTypeLookup["101"] = "In Service";
+    messageTypeLookup["100"] = "Returning";
+    messageTypeLookup["011"] = "Committed";
+    messageTypeLookup["010"] = "Special";
+    messageTypeLookup["001"] = "Priority";
+    messageTypeLookup["000"] = "Emergency";
```
ooglek, line 1048: Is there a more compact way to define a hash???

f4exb: To initialize a QHash with a literal definition you would do something like this:
```c++
	QHash<QString, QString> messageTypeLookup = {
		{"011", "Test"},
		{"001", "Hello"}
	};
```
Not necessarily more compact but somehow more elegant

ooglek:
I ran things overnight and had no issues, about 775 packets Demodded, and no errors or crashes in the APRS Feature. 

Argh, the polarity of the lat/lon are not right. I'll dig on that.

Screenshot of the AX.25 Packet decoded:

AX.25 APRS Packet decoded

Screenshot of the APRS Packet in Mic-E decoded:

APRS Mic-E Decoded

ooglek:
OK found the issue with Lat/Lon direction, was using the wrong letters. Fixed! It's a good first round.

I'll consider adding support for the APRS Mic-E Status Text support, which encodes the type of radio in to the Status Message, in the near future. Let me know if there's any further feedback or requested changes to follow standards! I'm still learning the finer points of C++, I'm only 2 days in. ;-)

ooglek:
Woot! Thrilled to contribute. LMK if there are any other APRS improvements needed! I hope to pick up and finish the APRS Mic-E Status Text support soon. Cheers and thanks for the help and guidance, @f4exb and @srcejon

PR #1115, opened 2022-01-26
Added support for LIB_SUFFIX (e.g. for lib64 on Fedora)

yarda:
On some distro's (e.g. Fedora, CentOS, ...) the 'lib64' directory is used for the 64 bit architecture dependent code instead of the 'lib'.


yarda:
FYI I am trying to get the sdrangel into Fedora, Fedora ticket:
https://bugzilla.redhat.com/bugzilla/show_bug.cgi?id=2045924

PR #1178, opened 2022-03-18
Add FeatureLayout - to try to better use space in Feature window

srcejon:
This patch adds a QLayout specifically for the Feature window, to try to better use the available space.

I think there are two problems with the current layout used in the Feature windows:

- It doesn't make efficient use of all available space

image

- Expandable widgets such as the Map, AIS or Radiosonde widgets can't be resized by the user - (the space is just shared equally).

image

The new FeatureLayout tries to address this, given that the widgets are basically in two categories: Fixed size like AFC, Jogdial, GS232 Controller etc and expandable like Map, AIS, or RadioSonde.

When the layout is set to horizontal, on the left, columns are created with the fixed widgets, then on the right the expandable widgets are placed in a QSplitter, which allows them to be resized. This results in:

image

This could possibly be made to look a bit nicer if we resize the fixed width widgets so they all have the same width.

For short / wide horizontal layouts, multiple columns will be used:

image

And here are the same widgets, having been resized by with the splitter:

image

When the layout is vertical, the fixed widgets are in rows, instead of columns, with the expandable widgets stacked in a vertical QSplitter:

image

Let me know what you think.

### DIFF
--- a/plugins/feature/aprs/aprsgui.ui
+++ b/plugins/feature/aprs/aprsgui.ui
@@ -11,7 +11,7 @@
    </rect>
   </property>
   <property name="sizePolicy">
-   <sizepolicy hsizetype="Expanding" vsizetype="Preferred">
+   <sizepolicy hsizetype="Expanding" vsizetype="Expanding">
     <horstretch>0</horstretch>
     <verstretch>0</verstretch>
    </sizepolicy>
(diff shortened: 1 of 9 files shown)


f4exb:
Why not it gives some flexibility in the display. However this will probably have to change when implementing a workspace concept (v7) hopefully....

PR #1205, opened 2022-04-19
V7

loewal:
Hello Edouard,

v7 is running fine on my iMac and Ubuntu Impish.
I have a question on 'deviceset presets':

When I open this window and apply a certain preset, that window is then closed.
Selecting a diiferent preset requires to open it again.
It's no big deal, but would be quite covenient:
Would it be possible to have that window open at all times? Or do you want to limit the number of open windows (and this was done 'on purpose')?


loewal:
> 

I'm sorry, but I noticed that this was 'fixed' in this alpha-version.
Sorry!

f4exb:
Glad to know this alpha is working for you but there is still some stuff to be done before merge. However I can keep this PR open in the meantime.

f4exb:
Conflicts to be solved later

PR #1215, opened 2022-04-25
Restore cursor when move from border to child widget

srcejon:
There's currently a small issue with the cursor not being restored after hovering over the border (which changes it to the resize cursor) then being moved immediately over a child widget. In this situation, the FramelessWindowResizer doesn't get a MouseMoveEvent.

This patch fixes this, by enabling mouse tracking on all child widgets.

A small change is also made to GLSpectrum, to enable propagation of MouseMoveEvents to the parent widget.


f4exb:
Great! I also noticed the cursor thing but wasn't able to figure out the root cause.

PR #1245, opened 2022-05-15
Fix typing errors in readme's

dforsi:
With this merge request I'm proposing to fix minor errors in readme files only because the tool `codespell` can also fix spelling in code, but it could also change names of functions and variables so it would need a longer list of words to ignore.

Let me know if some files should have been excluded.


f4exb:
This is only in .md files so it should be fine

PR #1262, opened 2022-05-29
Funcube dongle support on windows

srcejon:
This PR tries to enable Funcube dongle support on Windows.

It uses HIDAPI included as part of sdrangel-windows-libraries
fchhid and fcdlib libraries are now built on Windows.
fcdlib needed exports adding for global data, as cmake's WINDOWS_EXPORT_ALL_SYMBOLS only works on functions.

I don't have a dongle to test it - but at least the devices build now.


srcejon:
Is there something you or I need to do to link to the updated sdrangel windows libraries?

f4exb:
```
ninja: error: '../external/windows/hidapi/x64/hidapi.lib', needed by 'bin/fcdhid.dll', missing and no known rule to make it
```
As the "external" folder is a git submodule maybe it is just the case that you didn't update submodules.

srcejon:
> Edit: Indeed in your branch I see external/windows at commit 2d49813 but it should be commit 19a3329. If you update on your side I suppose this would become part of the PR and be merged accordingly.

Thanks. Seems I needed to do:

    cd sdrangel
    git submodule update --remote
    git add external/windows
    git commit

PR #1270, opened 2022-06-01
NoiseFigure plugin: Fix #1269 and #1268

### LINKED ISSUE #1268: Noise Figure Plugin - Can't run PSU control script
Problem present on Windows 10, SDRAngel 6.18.0 and Windows 7, SDRAngel 7.2.1

In the Noise Figure Plugin, I would like to use the 'Programs / scripts' function to control my PSU that powers my noise source. I don't have a VISA / SCPI PSU. Documentation says I should be able to use either control method.

There are default values in the 'VISA / SCPI' settings (that seem to match the authors test setup), which I cannot edit/delete as I don't have any VISA libraries installed.

When I try and run the Noise Figure measurement I get an error dialog saying 'Failed to open VISA device ....' and the Noise Figure measurement is terminated.

I can work around this issue by installing ~GigaByte of National Instruments VISA drivers, this then allows me to clear the default values in the 'VISA /SCPI' settings box. I then believe the Noise Figure Plugin is using my PSU control scripts. However, I can't verify this yet as I have come across another issue with the plugin.

Possible solutions

The plugin detects whether a VISA library is installed, In the case when there is no VISA library, any default settings in the 'VISA / SCPI' settings should be ignored or cleared.

I guess another solution would be to get rid of the default 'VISA / SCPI' settings in the code and leave them blank?

### LINKED ISSUE #1269: Noise Figure Plugin - won't run
Problem present on Windows 7, SDRAngel 7.2.1

After performing workaround to allow running of PSU control script (instead of VISA / SCPI PSU control), as mentioned in #1268, the Noise Figure Plugin doesn't appear to step through the frequencies and perform the Noise Figure measurement when pressing the 'play' button. 

The power meter in the Noise Figure plugin appears to be working correctly and responds appropriately to changes in receiver gain. However, the radio does not retune. I'm using the default frequency sweep settings (430 MHz to 440MHz in 3 steps).

I'm running Windows and I'm assuming that the PSU control script can be a batch file?

I have limited debug experience/capability but if I enable logging and try to start the Noise Figure measurement, I get the following log message, which may be relevant:

2022-05-31 11:59:15.568 (W) QMetaObject::invokeMethod: No such method ChannelAPI::nextState()

I get the same issue with both my SDR's, ADALM-PLUTO and SDRPlay.

Eventually, I would like to do this test on a networked USRP but understand there are some issues at present talking to an IP based USRP?

srcejon:
Fixes #1268 by allowing VISA settings to be adjusted even if VISA library not installed.
Fixes #1269 by using C++ syntax for slot rather than SLOT() macro.

### DIFF
--- a/plugins/channelrx/noisefigure/readme.md
+++ b/plugins/channelrx/noisefigure/readme.md
@@ -90,7 +90,7 @@ You can choose between linear and barycentric rational interpolation, and the di
 <h3>13: Open Noise Source Control dialog</h3>
 
 Opens the noise source control dialog, to allow setting how the plugin turns the power to the noise source off and on. Two control methods are supported: A program or script can be run to turn the power on or off,
-or the VISA library (if installed) can be used to send SCPI commands to a programmable power supply or other test equipment. If a VISA library is not found, the VISA and SCPI fields will be disabled.
+or the VISA library (if installed) can be used to send SCPI commands to a programmable power supply or other test equipment.
 
 The delay setting determines how long after the noise source's power is turned on or off, before the noise figure measurement starts. This should be long enough to allow the noise source output to settle, but
 not too long so that tests over a large number of frequencies take a long time to run. The ideal value will vary with the noise source and power supply.


Bob-Sellers:
Thank you for your speedy response and for an impressive piece of software.

I will test it as soon as a Windows release is available.

Bob-Sellers:
Pulled down the build artifact that hopefully resolves #1271 , (sdrangel-7.3.0-25-g6b0489021-win64.exe)

I have first tested it for Noise Figure Plugin functionality.

I can confirm that I can now edit/delete VISA/SCPI settings.

My PSU is out of action at the moment, but I produced Windows batch files that wrote text to a file, that temporarily substitute the 'Power on' and 'Power off' scripts.

With an ADALM-PLUTO I can now run Noise Figure measurements and sweep frequency and gain, the correct number of Power On and Power Off commands seem to have been issued by inspection of my test text file.

My Noise Figure results at the moment are clearly nonsense due to my non functioning noise source, but SDRAngel seems to be performing well!

PR #1291, opened 2022-06-20
3D Spectrogram

srcejon:
SDRangel already uses OpenGL, so couldn't resist :)

Demo here: https://www.youtube.com/watch?v=fCKjjDsE4S8

A couple of notes:

- To improve 3D quality, I have added an option for multisampling anti-aliasing (MSAA) to the spectrum window. This also results in the histogram being anti-aliased:

image

It can be enabled/disabled by the new Preferences > Graphics menu. Default is off - so it should look no different unless you switch it on. Note that the option only takes effect when the window is created, so you may need to restart SDRangel to see a difference.

There were a few problems getting this working on the Mac. Until now, all SDRangel GLSL shader code appeared to be using 2.x syntax. Unfortunately, on the Mac, when requesting an OpenGL 3.3+ context (which is needed for the Shaded rendering in the 3D Spectrogram and also the Map feature), because the Mac only supports the OpenGL 3 Core profile, none of the 2.x shaders work (Any deprecated features are not implemented in the Core profile). Also, in the 3 Core profile, VAOs are required, but weren't currently used. So I have updated all of the shaders to have two versions: one using OpenGL 2 syntax for OpenGL < 3.3 and the other with version 330 shaders for OpenGL 3.3+. Tested on:

Windows - opengl 4.6
Windows - opengl 3.0 (ANGLE)
Ubuntu 20.04 - opengl 3.3
Ubuntu 22.04 - opengl 4.1
Mac - opengl 4.1

I don't have a Pi to test on - but hopefully should work. If not, let me know!


f4exb:
I checked the video and it looks nice indeed. Not sure the choice of colormap also applies in 2D. I think that would be useful too but we can check later.

srcejon:
Yes, I didn't try to change any of the 2D stuff, so it just applies to 3D currently (the combo box is hidden in 2D mode). But should be trivial to add.

f4exb:
Interesting... particularly the gradient ones as they can give a visual differentiation of the magnitudes.

f4exb:
On a different note we may need to split the first block of controls in the GUI (there are currently 4 blocks in a flow layout that make a best fit of the GLSpectrum GUI controls space depending on the overall spectrum window shape without breaking at illogical places). I can take care of this later.

Issue #1284, opened 2022-06-10
USRP X310 - failed to open device


Bob-Sellers:
Windows 10, with SDRangel 7.3.1-19-g6029d3089

Ettus X310 fitted with 2 x TwinRX daughterboards (ie radio has 4 RX channels and 0 TX channels)

I was wanting to perform some Noise Figure measurements on an Ettus X310 radio using SDRangel's excellent new support for networked USRP radios. However, I've come across a problem. I can't 'open' the radio.

The log reads as follows:

2022-06-10 14:18:44.243 (D) DSPDeviceSourceEngine::handleSynchronousMessages:  DSPAddBasebandSampleSink
2022-06-10 14:18:44.244 (D) MainWindow::sampleSourceCreate: deviceSetIndex: 0 deviceIndex: 6 hardwareId: "USRP" sequence: 0 id: "sdrangel.samplesource.usrp" serial: "type=x300,addr=192.168.10.2,fpga=HG,name=,serial=XXXXXXX,product=X310" displayedName: "[0:0] XXXXXXX"
2022-06-10 14:18:44.244 (D) USRPInput::suspendRxBuddies (0)
2022-06-10 14:18:44.244 (D) USRPInput::suspendTxBuddies (0)
2022-06-10 14:18:44.249 (D) USRPInput::openDevice: allocated SampleFifo
2022-06-10 14:18:44.249 (D) USRPInput::openDevice: open device here
2022-06-10 14:18:44.249 (D) DeviceUSRPParams::open: type=x300,addr=192.168.10.2,fpga=HG,name=,serial=XXXXXXX,product=X310
2022-06-10 14:18:47.484 (D) DeviceUSRPParams::open: exception:  LookupError: KeyError: Invalid TX channel: 0
2022-06-10 14:18:47.484 (C) USRPInput::openDevice: failed to open device
2022-06-10 14:18:47.484 (D) USRPInput::resumeTxBuddies (0)
2022-06-10 14:18:47.484 (D) USRPInput::resumeRxBuddies (0)
2022-06-10 14:18:47.491 (D) DSPDeviceSourceEngine::setSource
2022-06-10 14:18:47.492 (D) DSPDeviceSourceEngine::handleSynchronousMessages:  DSPSetSource
2022-06-10 14:18:47.492 (D) DSPDeviceSourceEngine::gotoIdle
2022-06-10 14:18:47.492 (D) DSPDeviceSourceEngine::handleSetSource: set USRPInput
[...]

PR #1293, opened 2022-06-20
USRP: Attempt to support radios without AGC

srcejon:
Additional fix for #1284.


Bob-Sellers:
I would be interested to test this, but it looks like the build has failed?

srcejon:
It should be in some of the other more recent builds, so try this: https://github.com/f4exb/sdrangel/suites/7046820635/artifacts/277584925

PR #1299, opened 2022-06-22
Spectrum fill/gradient

srcejon:
This PR adds support for drawing the spectrum as filled (including max hold) or with a gradient using the colour map. The colour map is now also applied to 2D waterfall.

image

image

Note, I've added a combo to select between Line / Fill / Gradient. However, you might want to change this logic so that it switches between Line / Fill / Gradient / Phosphor effect - as the phosphor effect doesn't work well with the filled spectrum. Icons might be nicer/more compact than words too - you do a better job of making those!

Any idea why the m_waterfallPalette is only 240 entries? I haven't tried to expand that - but all of the Color Maps I've added are 256 entries - so it possibly could be.

### DIFF
--- a/sdrbase/dsp/spectrumsettings.cpp
+++ b/sdrbase/dsp/spectrumsettings.cpp
@@ -66,7 +66,8 @@ void SpectrumSettings::resetToDefaults()
 	m_useCalibration = false;
 	m_calibrationInterpMode = CalibInterpLinear;
     m_3DSpectrogramStyle = Outline;
-    m_3DSpectrogramColorMap = "Angel";
+    m_colorMap = "Angel";
+    m_spectrumStyle = Line;
 }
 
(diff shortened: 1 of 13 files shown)


f4exb:
>  you do a better job of making those! 

Thanks! Yes don't worry I will re-arrange this and probably split the first line into sub sections as it is growing too big.

> Any idea why the m_waterfallPalette is only 240 entries

Well this is very old code probably from the early SDRangelove. I have to inspect the code to retrieve the logic (or not logic).

f4exb:
> the phosphor effect doesn't work well with the filled spectrum

I think it's OK (maybe a video would render the effects better...):
image

Note: I have started re-organizing the icons a little bit.

srcejon:
Yeah - having another look, some colour schemes work OK (if max hold is switched off). Looking good.

PR #1305, opened 2022-06-24
Spectrum: Allow frequency scrolling

srcejon:
This patch allows a user to adjust center frequency up or down in a main spectrum by:

- Holding down the middle mouse button and moving the mouse left or right
- Moving the mouse outside of the spectrum area when dragging a channel marker.

For me, this makes it much quicker and easier to move to frequencies that are just out of range. 

It works quite well on RTL SDR & USRP (for the latter, it's nearly as smooth as moving the channel marker) - but is a little slow on SDRplay.

Note, this works by GLSpectrum emitting a signal requesting a change of center frequency - The change of device frequency is handled by MainSpectrumGUI - so GLSpectrum doesn't need to know whether it is connected to a device or something else.

### DIFF
--- a/sdrgui/gui/glspectrum.cpp
+++ b/sdrgui/gui/glspectrum.cpp
@@ -92,6 +92,8 @@ GLSpectrum::GLSpectrum(QWidget* parent) :
     m_scaleZ3DSpectrogram(false),
     m_3DSpectrogramStyle(SpectrumSettings::Outline),
     m_colorMapName("Angel"),
+    m_scrollFrequency(false),
+    m_scrollStartCenterFreq(0),
     m_histogramBuffer(nullptr),
     m_histogram(nullptr),
     m_displayHistogram(true),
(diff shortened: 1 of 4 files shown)


srcejon:
On a different subject, did you mean to create the new waterfall.png and waterfall_3d.png icons so big? They're 480x480 pixels.

f4exb:
> On a different subject, did you mean to create the new waterfall.png and waterfall_3d.png icons so big? They're 480x480 pixels.

Looks like they were left at their original size when drawing them in Gimp.

f4exb:
To be used with care... in a multi channel configuration the change in the device frequency affects all channels. You may want this or not.

f4exb:
I think it is better if `MainWindow` takes care of device center frequency setting. Moreover it has access to the `DeviceAPI` object which can provide access to the virtual `setCenterFrequency` method of source and device sinks. This can provide support to a wider range of devices and maybe improve performance. I have corrected the code in that sense so that `MainSpectrumGUI` relays the signal of `GLSpectrum` adding the device set index for commodity and is later caught and processed in the `MainWindow`.

srcejon:
Ok, sounds good. Thanks for updating the docs!

PR #1309, opened 2022-06-24
SpectrumGUI: Add combo box to allow center frequency to be set to marker

srcejon:
Last one :)

This patch adds a combo box to the Spectrum GUI that displays all of the annotation markers - and when pressed, sets the center frequency accordingly, allowing you to quickly jump between pre-defined frequencies.

Combo box is only displayed if there are annotation markers.


f4exb:
I have added a patch over this so that the combo is also made visible only if the spectrum being displayed is the spectrum of a device (i.e. "main spectrum") else it doesn't make sense and you can have markers for any kind of spectrum. Also it updates the markers list after the markers dialog is called in order to take account any updates.

PR #1288, opened 2022-06-13
USRP: Don't call get_tx/rx methods if tx/rx channels is 0, as may result in KeyError exception

srcejon:
First attempt to fix #1284

PR #1315, opened 2022-06-29
USRP: Catch all exceptions from set_rx_agc

srcejon:
For #1284 and #1288


Bob-Sellers:
Using the build artifact from this, ie: sdrangel-7.4.0-7-g486b74b31-win64

This seems to be working much better.

- When the sample rate is changed, the sample rate now snaps to the nearest available rate.
- The spectrum frequency axis scales correctly when the sample rate is changed
- sensible looking frequencies are selected
- on previous versions the master clock rate was not reported in the 'Device' window, it just showed up as '-', it now shows up as '200M'

I'll continue to play, but all looks good at the moment. Well done!

Issue #1341, opened 2022-07-16
DATV LDPC decoding question


DL9UFB:
It is possibele to implement LDPC in the Windows Version for receiving DATV on QO-100?

f4exb:
`ldpctool` is a Linux-centric external program. So... no.

srcejon:
> `ldpctool` is a Linux-centric external program. So... no.

Could it not be modified to build as a library? I've only had a quick glance, but it seems fairly OS independent apart from ldpc_tool.cpp, which doesn't seem that big, so could potentially be adapted.

srcejon:
With a just few changes ldpc_tool can build on Windows - the process management in leansdr/dvbs2 looks a little trickier though - but pipes to subprocesses can be made to work.

f4exb:
Or use QThread...

srcejon:
To clarify, I meant I can build ldpctool as a .exe on Windows - and in that case, pipes are needed for I/O to the child processes stdin/stdout. I've just written the code for this, but not working yet.

If it were to be rewritten as a library - then yes, QThread makes sense.

f4exb:
I see but aren't pipes used already? Cf https://github.com/f4exb/sdrangel/blob/master/plugins/channelrx/demoddatv/leansdr/dvbs2.h#L3451

f4exb:
Implemented in v7.6.0

PR #1345, opened 2022-07-18
DATV Demod: Add support for LDPC on Windows

srcejon:
This PR adds support for LDPC on Windows as per #1341. It does this by implementing ldpctool as a Qt worker thread instead of an external process. This can be used on Linux as well, avoiding the need for the ldpctool binary.

However, I've left the lpdctool path in the GUI & settings in case you wanted to make it optional which to use - but it could be removed, if you think this is the way to go. 

I don't really use this plugin, so have only done some basic testing on Linux and Windows. I think there's a DATV demod problem on Windows that can occasionally cause a crash, that isn't related to this change. Needs more investigation though.

### DIFF
--- a/plugins/channelrx/demoddatv/CMakeLists.txt
+++ b/plugins/channelrx/demoddatv/CMakeLists.txt
@@ -25,6 +25,7 @@ set(datv_SOURCES
 
 set(ldpc_SOURCES
     ldpctool/tables_handler.cpp
+    ldpctool/ldpcworker.cpp
 )
 
 set(datv_HEADERS
@@ -55,6 +56,7 @@ set(ldpc_HEADERS
     ldpctool/dvb_s2_tables.h
     ldpctool/dvb_s2x_tables.h
     ldpctool/dvb_t2_tables.h
+    ldpctool/ldpcworker.h
 )
 
 include_directories(
@@ -69,16 +71,10 @@ include_directories(
 set(TARGET_NAME demoddatv)
 set(INSTALL_FOLDER ${INSTALL_PLUGINS_DIR})
 
-if (LINUX)
-    add_library(${TARGET_NAME} SHARED
-        ${datv_SOURCES}
-        ${ldpc_SOURCES}
-    )
-else()
-    add_library(${TARGET_NAME} SHARED
-        ${datv_SOURCES}
-    )
-endif()
+add_library(${TARGET_NAME} SHARED
+    ${datv_SOURCES}
+    ${ldpc_SOURCES}
+)
 
 target_link_libraries(${TARGET_NAME}
     Qt5::Core
@@ -94,13 +90,11 @@ target_link_libraries(${TARGET_NAME}
     ${SWRESAMPLE_LIBRARIES}
 )
 
-if (LINUX)
-    add_executable(ldpctool
-        ldpctool/ldpc_tool.cpp
-        ldpctool/tables_handler.cpp
-    )
-    install(TARGETS ldpctool DESTINATION ${INSTALL_BIN_DIR})
-endif()
+add_executable(ldpctool
+    ldpctool/ldpc_tool.cpp
+    ldpctool/tables_handler.cpp
+)
+install(TARGETS ldpctool DESTINATION ${INSTALL_BIN_DIR})
 
 if(FFMPEG_EXTERNAL)
     add_dependencies(${TARGET_NAME} ffmpeg)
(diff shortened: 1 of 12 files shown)


f4exb:
Sorry this is not working. With the original soft LDPC the decoding is fluid but gets bursty with this version. I think the main thread is affected as even the GUI is moving by bursts. I think this means that the multiprocessing is not working effectively.

srcejon:
Hmm. One possible cause is that I made receive_frame blocking. I don't really understand the original code:

    void receive_frame(const helper_job *job)
    {
        // Read corrected frame from helper
        const s2_pls *pls = &job->pls;
        int iosize = (pls->framebits() / 8) * sizeof(ldpc_buf[0]);
        int nr = read(job->h->fd_rx, ldpc_buf, iosize);

        if (nr < 0)
        {
            if (errno != EAGAIN) { // if no data then try again next time
                fatal("s2_fecdec_helper::receive_frame read error");
            }
        }
        else if (nr != iosize)
        {
            fprintf(stderr, "s2_fecdec_helper::receive_frame: %d bytes read vs %d\n", nr, iosize);
        }

        --job->h->b_out;
        // Decode BCH.
        const modcod_info *mcinfo = check_modcod(job->pls.modcod);
        const fec_info *fi = &fec_infos[job->pls.sf][mcinfo->rate];
        uint8_t *hardbytes = softbytes_harden(ldpc_buf, fi->kldpc / 8, bch_buf);

It looks like the read file descriptor (job->h->fd_rx) is made non-blocking (just after it is opened), but if it doesn't block, I don't understand what the above code does, as it looks like it uses ldpc_buf to decode, even though nothing was read in to it.

Is that something you wrote? If not, what was the source of leansdr? https://github.com/pabr/leansdr seems to be an older version.

srcejon:
If you want to experiment, in the new receive_frame, where you have:

        QByteArray data = job->h->m_worker->data();

Which performs the blocking read of data, you could change this to something like:

    if (job->h->m_worker->dataAvailable())
        QByteArray data = job->h->m_worker->data();
   else 
       skip it

But I don't see how that would work as the call to receive_frame removes the job:

    receive_frame(jobs.get())

f4exb:
Yes this is something I don't fully understand either. Probably the reason why I did not want to change the code and gave up on Windows. The active branch is not master but "work": https://github.com/pabr/leansdr/tree/work

f4exb:
To illustrate what happens:

With original  ldpctool:
https://user-images.githubusercontent.com/6192319/179631208-84017967-1c53-4cd7-8881-ac3c2129ba8c.mp4

With the fix_1341 branch changes:
https://user-images.githubusercontent.com/6192319/179631388-e1b3df91-6eac-4678-b4dc-464fabc84074.mp4

f4exb:
Well... this seems to work:
```diff
--- a/plugins/channelrx/demoddatv/leansdr/dvbs2.h
+++ b/plugins/channelrx/demoddatv/leansdr/dvbs2.h
@@ -3829,8 +3829,11 @@ struct s2_fecdec_helper : runnable
 
         // Blocking read - do we need to return faster?
         // If so, call m_worker->dataAvailable()
-        QByteArray data = job->h->m_worker->data();
-        memcpy(ldpc_buf, data.data(), data.size());
+        if (job->h->m_worker->dataAvailable())
+        {
+            QByteArray data = job->h->m_worker->data();
+            memcpy(ldpc_buf, data.data(), data.size());
+        }
 
         --job->h->b_out;
         // Decode BCH.
```

srcejon:
Ok, that makes it more like the original code - but I don't understand it!

I just tried here do back-to-back stream with a Lime with similar settings, and I don't see the problem:

image

Possibly CPU performance related? But if that change works, lets go with it.

f4exb:
You have a much larger MER and CNR than on my reception of the QO-100 beacon so the LDPC stuff is not solicited very much. I think this is the explanation. It is very important that reads are not blocking else this compromises the effective multi processing which is critical to deliver the blocks on time.

So I think we can go with the "non blocking read" fix. Moreover it seems to work better than the original may be due to the fact that there is probably less overhead than when calling an external program.

f4exb:
It still works at quite low MER:
image

f4exb:
This works as well and is probably more efficient:
```diff
--- a/plugins/channelrx/demoddatv/leansdr/dvbs2.h
+++ b/plugins/channelrx/demoddatv/leansdr/dvbs2.h
@@ -3829,8 +3829,15 @@ struct s2_fecdec_helper : runnable
 
         // Blocking read - do we need to return faster?
         // If so, call m_worker->dataAvailable()
-        QByteArray data = job->h->m_worker->data();
-        memcpy(ldpc_buf, data.data(), data.size());
+        if (job->h->m_worker->dataAvailable())
+        {
+            QByteArray data = job->h->m_worker->data();
+            memcpy(ldpc_buf, data.data(), data.size());
+        }
+        else
+        {
+            return;
+        }
 
         --job->h->b_out;
         // Decode BCH.
```

Issue #1329, opened 2022-07-08
Make settings assignments atomic


srcejon:
I think there are a couple of multi-threading problems with webapiSettingsPutPatch in devices and channels.

Using AirspyHF code as an example, but the same applies elsewhere:

        int AirspyHFInput::webapiSettingsPutPatch(
                        bool force,
                        const QStringList& deviceSettingsKeys,
                        SWGSDRangel::SWGDeviceSettings& response, // query + response
                        QString& errorMessage)
        {
            (void) errorMessage;
            AirspyHFSettings settings = m_settings;
            webapiUpdateDeviceSettings(settings, deviceSettingsKeys, response);
        
            MsgConfigureAirspyHF *msg = MsgConfigureAirspyHF::create(settings, force);
            m_inputMessageQueue.push(msg);
        
            if (m_guiMessageQueue) // forward to GUI if any
            {
                MsgConfigureAirspyHF *msgToGUI = MsgConfigureAirspyHF::create(settings, force);
                m_guiMessageQueue->push(msgToGUI);
            }
        
            webapiFormatDeviceSettings(response, settings);
            return 200;
        }

Possible problems include:

- No mutex is used. This can be problematic  for settings = m_settings, as this assignment isn't atomic - so if m_settings is being simultaneously written in applySettings(), settings may contain a mixture of old and new values.
- If webapiSettingsPutPatch is called quickly twice in succession, handling of the Configure message pushed to the queue may not have been handled by the time of the second call - and thus settings will contain old settings. This can result in some settings being lost, if not all keys are specified in each call. (E.g. if there's one patch to log2Decim followed by another patch to devSampleRate, the setting for log2Decim can be lost).
- Similarly, if webapiSettingsGet is called quickly after webapiSettingsPutPatch, old data may be returned for the get.

srcejon:
> For the first bullet this is very unlikely to happen and not worth a mutex.

There is already a mutex which I think should be suitable, just a case of adding:

    QMutexLocker mutexLocker(&m_mutex);

> For the middle bullet the solution is to return 204 that just says the messages have been processed. This is essentially asynchronous and we can't wait for the messages to be processed before returning the reply.

It's reasonable that processing is async, but how is a client to know when it is safe to send the next patch request?

> Also for the last bullet the process being essentially asynchronous this is something to cope with.

How though? Keep calling get until the patched value is returned?

f4exb:
This is a one way communication and how REST APIs work and one has to live with it.

[...]

Issue #456, opened 2019-12-15
RTL TCP


gaspar1987:
Hello, I'd like to know how we use your tcp.
To activate the software.
Thank you for giving me more information on this to help me solve the problem. Thank you.
![image](https://user-images.githubusercontent.com/13591469/70864997-98d68480-1f58-11ea-8caa-ce2773e487b2.png)

f4exb:
You cannot use RTL-SDR via TCP it has to be connected to your system via USB. There are no plans to have RTL-TCP supported... period.

dterweij:
Oh that is pitty. My RTL_SDR is plugged in at the PC located in the attic. Close to the antenna (3 meters of cable). 90% of all tools I do use is over RTL_TCP connection to my PC downstairs.  Please try to plan it...  I am not planning to move the equipment from attic to downstairs. Period :)

f4exb:
This issue was closed for two good reasons:
  - Streaming real time data at relatively fast rate over TCP is a bad idea so I will not promote it
  - SDRangel has its own options for remote operation that use UDP and FEC redundancy for protection against block loss. Please check these plugins:
    - https://github.com/f4exb/sdrangel/blob/master/plugins/channelrx/remotesink/readme.md (sender side)
    - https://github.com/f4exb/sdrangel/blob/master/plugins/samplesource/remoteinput/readme.md (receiver side)

In addition there is a server flavor of SDRangel that lets you run it on a RPi4 (for example) as a server. There are Docker images in Docker hub to just do that: https://hub.docker.com/r/f4exb06/sdrangelsrv The sdrangel-docker ancillary project: https://github.com/f4exb/sdrangel-docker contains more info about running SDRangel in Docker.

So really no reason to support RTL TCP.

synthead:
I would really like to see rtl_tcp support, even if there are other options :+1:

betzburger:
It looks like sdrangel is a really nice program with many features and I would love to try it. My Raspberry 4 is in the attic and the only way to use it by now is to send the data with rtl_tcp to the network and take the data back from the network with gqrx. This is how I do at the moment. For me (sorry) it seems it is super complicated to do the same with sdrangel / sdrangelsrv. At least I can't find any easy to follow step-by-step instruction. If there is some it would be fantastic.

icedream:
[...]

PR #1350, opened 2022-07-19
Remote TCP Input and Remote TCP Sink Plugins

srcejon:
This PR adds two new plugins: Remote TCP Input and Remote TCP Sink. They are similar to the current Remote Input and Remote Sink plugins, but use TCP rather than UDP.

Two protocols are supported:

- RTL0 - which allows the Remote TCP Input plugin to connect to rtl_tcp and rsp_tcp servers. It also supports Remote TCP Sink appearing as though it were rtl_tcp, allowing applications that use rtl_tcp to connect to a wider range of SDRs. This protocol is limited to 8-bit IQ data.
- SDRA - a slightly enhanced version of RTL0, that supports a wider range of device and channel settings and 16/24/32 bit IQ data, as well as the ability to send the remote device's settings when connecting so they can be displayed in the GUI. This protocol is recommend for when connecting Remote TCP Input to Remote TCP Sink.


srcejon:
The Remote TCP Sink plugin is what showed the problem reported in https://github.com/f4exb/sdrangel/issues/1329 - so device settings on something like SDRplay might not all get set properly for the initial connection.

f4exb:
Implements #456

PR #1352, opened 2022-07-19
Update rtlsdr external project to latest version

### LINKED ISSUE #1351: Exception when closing RTLSDR device
When closing an RTLSDR device on Windows 10, if acquisition has started at least once, I see an exception in libusb:

image

image

srcejon:
Appears to fix #1351 for me.

(Existing version appears to be from Jul 2019)


srcejon:
Working for me, as I don't have pkg-config installed, but in this build it looks like it is being picked up from: D:/a/sdrangel/sdrangel/external/windows/pkg-config-lite/bin/pkg-config.exe

When pkg-config is available, the latest rtlsdr CMake ignores LIBUSB_INCLUDE_DIRS and it wants a libusb .pc file, which we don't currently have in sdrangel-windows-libraries.

PR #1374, opened 2022-08-04
MacOS build fixes

srcejon:
A couple of fixes to M17 and DATV Demod so that it builds on MacOS Mohave.

### DIFF
--- a/modemm17/Correlator.h
+++ b/modemm17/Correlator.h
@@ -8,6 +8,7 @@
 #include <algorithm>
 #include <array>
 #include <cstdint>
+#include <cmath>
 #include <cstddef>
 #include <type_traits>
 #include <tuple>
--- a/plugins/channelrx/demoddatv/datvdemodsink.cpp
+++ b/plugins/channelrx/demoddatv/datvdemodsink.cpp
@@ -397,9 +397,11 @@ void DATVDemodSink::CleanUpDATVFramework()
         delete (leansdr::s2_fecdec<bool, leansdr::hard_sb>*) r_fecdec;
     }
 
+#if 0
     if (r_fecdecsoft != nullptr) {
         delete (leansdr::s2_fecdec_soft<leansdr::llr_t,leansdr::llr_sb>*) r_fecdecsoft;
     }
+#endif
 
     if (r_fecdechelper != nullptr) {
         delete (leansdr::s2_fecdec_helper<leansdr::llr_t,leansdr::llr_sb>*) r_fecdechelper;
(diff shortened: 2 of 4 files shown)


f4exb:
So if I get it correctly we will have soft LDPC decoding only for Linux (we drop it for Windows as well) ?

srcejon:
No - I've just defined out some unused code as it doesn't compile on Mac. But yes, it is a bit confusing and could do with a further tidy up.

In datvdemodsink.cpp, before any of these LDPC changes, there was some code that was defined out:

    #if 0
        // Doesn't work...
        // Soft LDPC decoder mode.
        // Deinterleave into soft bits.
        p_fecframes = new leansdr::pipebuf<leansdr::fecframe<leansdr::llr_sb> >(m_objScheduler, "FEC frames", BUF_FRAMES);
        p_s2_deinterleaver = new leansdr::s2_deinterleaver<leansdr::llr_ss, leansdr::llr_sb>(
            m_objScheduler,
            *(leansdr::pipebuf< leansdr::plslot<leansdr::llr_ss> > *) p_slots_dvbs2,
            *(leansdr::pipebuf< leansdr::fecframe<leansdr::llr_sb> > * ) p_fecframes
        );
        r_fecdecsoft = new leansdr::s2_fecdec_soft<leansdr::llr_t,leansdr::llr_sb>(
            m_objScheduler, *(leansdr::pipebuf< leansdr::fecframe<leansdr::llr_sb> > * ) p_fecframes,
            *(leansdr::pipebuf<leansdr::bbframe> *) p_bbframes,
            m_modcodModulation < 0 ? 0 : m_modcodModulation,
            true, 5,
            p_vbitcount,
            p_verrcount
        );
    #else
        // External LDPC decoder mode.
        // Deinterleave into soft bits.
       ....

The bit of code I added the #ifdef LINUX for in this commit is only used inside the #if 0 here - so both bits of code could potentially be removed completely, unless you see a future need for them.

The "External LDPC decoder mode" above is the code that is currently used, even though the decoder is no longer external, so that could do with a bit of a tidy up as well. 

We could also remove the USE_LDPC_TOOL #defined code as well. Depends if you just want to move forward with only the one implementation.

f4exb:
OK I probably skipped some details indeed the original code is not very readable. 

> We could also remove the USE_LDPC_TOOL #defined code as well. Depends if you just want to move forward with only the one implementation.

In fact I have been testing the "internal" implementation and it works at least as well as with the "external" implementation. There might even be some gain in processing since there is not the overhead of spawning a full process. SO I think that eventually yes the old code could be removed.

PR #1404, opened 2022-09-06
MSVC M17 FIX

Paulo-D2000:
Hi, I fixed the msvc errors and now it builds on msvc :)
 I did sent you one email yesterday about fixing the Tx v7.6.0, after that the Rx was fixed too and released under my branch...
This is the 7.6.3 (latest) with the fix too.. 
Fix details:
*Enabled the modm17 & demodm17 build on cmakelists
*Changed the template naming as msvc can go crazy with repeated typenames...
*Redudant header includion on m17demodplugin removed


f4exb:
> Changed the template naming as msvc can go crazy with repeated typenames...

Well spotted. That's quite strange indeed.

PR #1405, opened 2022-09-07
Fix c++17 compilation with MSVC

srcejon:
set(CMAKE_CXX_STANDARD 17) currently fails on Windows with an error in Windows SDK headers:

    C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\shared\rpcndr.h(192): error C2872: 'byte': ambiguous symbol
    C:\Program Files (x86)\Windows Kits\10\include\10.0.19041.0\shared\rpcndr.h(191): note: could be 'unsigned char byte'

This patch implements suggested workaround here:

    https://developercommunity.visualstudio.com/t/error-c2872-byte-ambiguous-symbol/93889#T-N859630

Note, the patch doesn't set CMAKE_CXX_STANDARD to 17 - it's left at 14.


kasper93:
Proper fix would be probably to remove any and all `using namespace std;` which are before including Windows SDK headers. It is not nice to do so. Generally bringing namespaces into global scope should be done with care and at most at translation unit level.

srcejon:
> Proper fix would be probably to remove any and all `using namespace std;` which are before including Windows SDK headers. It is not nice to do so. Generally bringing namespaces into global scope should be done with care and at most at translation unit level.

Yep, sounds fair. Have pushed updated patch.

Issue #1409, opened 2022-09-15
APRS source does not see MIMO sourced packet demodulator


audriusb:
If i open up my LimeSDR with MIMO and add a pkt demodulator, this does not appear in APRS sources list.
If opened as single RX device: everything is fine. 

(Though i get segfaults, time to time on simple RX with limesdr, but i blame ubuntu 22.04 for that, since I compiled sdrangel myself and this does not have any issues with rtlsdr). 

Version: 7.6.3

srcejon:
When APRS::scanAvailableChannels() iterates through deviceSets, it only looks for DSPDeviceSourceEngine, not DSPDeviceMIMOEngine.

srcejon:
The above patch should fix the Packet Demod not appearing in the sources list.

I can't receive APRS right now, so do not know if it will work completely - so please let me know if you still have a problem.

audriusb:
Works like a charm. Thanks for the speed. 
![APRSTest](https://user-images.githubusercontent.com/29307678/190569109-e93e2a89-b335-4543-8de5-bbe3ddac1b01.png)

PR #1410, opened 2022-09-15
#1409 - Detect Packet Demods in MIMO device sets in ARPS Feature

srcejon:
This PR for #1409 enables the APRS Feature to detect Packet Demods in MIMO device sets.


audriusb:
That was fast. I'll give it a try in a few hours. Thanks.

audriusb:
Odd behaviour that after compiling I had to recreate my config file. Anyway, tested it: works fine.

PR #1413, opened 2022-09-15
Replace deprecated QMutex(Recursive) with QRecursiveMutex

pinkavaj:
The goal here is to reduce noise during compiling.

The `QMutex m_mutex(QMutex::Recursive)` construct is deprecated and should be replaced by `QRecursiveMutex`.

### DIFF
--- a/plugins/channelrx/sigmffilesink/sigmffilesinkbaseband.cpp
+++ b/plugins/channelrx/sigmffilesink/sigmffilesinkbaseband.cpp
@@ -33,8 +33,7 @@ SigMFFileSinkBaseband::SigMFFileSinkBaseband() :
     m_channelizer(&m_sink),
     m_specMax(0),
     m_squelchLevel(0),
-    m_squelchOpen(false),
-    m_mutex(QMutex::Recursive)
+    m_squelchOpen(false)
 {
     qDebug("SigMFFileSinkBaseband::SigMFFileSinkBaseband");
     m_sampleFifo.setSize(SampleSinkFifo::getSizePolicy(48000));
--- a/logging/logger.cpp
+++ b/logging/logger.cpp
@@ -7,6 +7,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <QMutex>
+#include <QRecursiveMutex>
 #include <QDateTime>
 #include <QThread>
 #include <QObject>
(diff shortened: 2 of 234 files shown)


f4exb:
Only since Qt 5.14 and we still need to be able to compile for Qt 5.12 so this is a breaking change

f4exb:
Note that the main thing holding us back to Ubuntu 20.04 and therefore Qt 5.12.8 is the Appveyor image for CI. However it seems possible to use Qt 5.15.2 inside this image but this needs some investigation and it is not ready right now.

plugins/channelrx/sigmffilesink/sigmffilesinkbaseband.cpp
```diff
@@ -34,7 +34,6 @@ SigMFFileSinkBaseband::SigMFFileSinkBaseband() :
     m_specMax(0),
     m_squelchLevel(0),
     m_squelchOpen(false),
```
f4exb, line 36: Extra comma here

pinkavaj: Thx, fidex

pinkavaj: But the `../logging/logger.h:13:10: fatal error: QRecursiveMutex: No such file or directory` is probably issue with `Qt 5.12` right?

pinkavaj: > Note that the main thing holding us back to Ubuntu 20.04 and therefore Qt 5.12.8 is the Appveyor image for CI. However it seems possible to use Qt 5.15.2 inside this image but this needs some investigation and it is not ready right now.

What do you suggest, I can hold till the Qt version issue is resolved. Should I close this MR for now?

f4exb: Indeed we will not try to make it work for Qt 5.12 and also later versions with #define(s). So in any case it should wait for Qt upgrade whether it remains open and unmerged or closed and submitted later. I know keeping Qt 5.12 compatibility is a pain (another PR is failing to compile because of this) and upgrade should take place one way or the other and the sooner the better.

f4exb: https://github.com/f4exb/sdrangel/issues/1416 opened

f4exb: Now building for Ubuntu 22.04 with Qt 5.15

PR #1414, opened 2022-09-16
Remote Control Feature

srcejon:
This PR adds a Remote Control Feature, for remotely controlling and monitoring Smart Home / IoT devices such as plugs and switches, as well as VISA programmable test equipment, such as benchtop power supplies and spectrum analyzers.

This is so users can remotely control power to SDRs, power amplifiers, rotator controllers and the like from within SDRangel. (I'm trying to cut down on electricity usage!)

The GUI is customizable so users can choose which controls and sensors to display, as well as plot charts of sensor data over time.

image


plugins/feature/remotecontrol/remotecontrolgui.cpp
```diff
@@ -0,0 +1,1174 @@
...
+            }
+            break;
+
+        case DeviceDiscoverer::LIST:
+            {
+                QComboBox *combo = new QComboBox();
+
+                combo->setToolTip("Set value for " + info->m_name);
+                combo->insertItems(0, info->m_values);
+                connect(combo, &QComboBox::textActivated,
```
f4exb, line 446: Compile error (Qt 5.12): error: ‘textActivated’ is not a member of ‘QComboBox’

Issue #1346, opened 2022-07-18
ASSERT failure in QCoreApplication::sendEvent: when closing Features


srcejon:
When building with the debug version of Qt libraries, assertions are enabled. It seems we get an assertion when several Features are closed. For example, AFC.

The assertions are of the form:

2022-07-18 17:20:10.316 (F) ASSERT failure in QCoreApplication::sendEvent: "Cannot send events to objects owned by a different thread. Current thread 0x0x21435329da0. Receiver '' (of type 'AFCWorker') was created in thread 0x0x21444945170", file kernel\qcoreapplication.cpp, line 558

This appears to be due to worker objects being deleted in the destructor of the Feature's main class:

    AFC::~AFC()
    ....
    delete m_worker;

and that the worker objects are moved to a different thread in the constructor:

    AFC::AFC
    ...
    m_worker = new AFCWorker(webAPIAdapterInterface);
    m_worker->moveToThread(&m_thread);

Note that we can't just do:

    m_worker->deleteLater();

As at this point, m_thread has already been stopped, so the worker's destructor never gets called.

We can't also do:

    m_worker->moveToThread(QThread::currentThread());
    delete m_worker;

As you can only push an object to a thread, not pull it (i.e. moveToThread needs to be executed in the content of m_thread)

The recommended approach https://wiki.qt.io/QThreads_general_usage seems to be connect &Worker::finished to &Worker::deleteLater and &QThread::finished to &QThread::deleteLater - however, that doesn't work here as SDRangel workers can be restarted.

One thing that seems to work is starting the thread then calling deleteLater()

    m_thread.start();
    m_worker->deleteLater();
    m_thread.quit();
    m_thread.wait();

As this results in the worker's destructor being called.

Any better ideas?

f4exb:
It is a bit strange to restart the thread just for this purpose. Instead the destructor code may be rewritten this way (this is the APRS feature because I had this one in my config):

```diff
--- a/plugins/feature/aprs/aprs.cpp
+++ b/plugins/feature/aprs/aprs.cpp
@@ -83,11 +83,14 @@ APRS::~APRS()
         &APRS::networkManagerFinished
     );
     delete m_networkManager;
+
     if (m_worker->isRunning()) {
-        stop();
+        m_worker->stopWork();
     }
 
-    delete m_worker;
+    m_worker->deleteLater();
+    m_thread.quit();
+    m_thread.wait();
 }
 
 void APRS::start()
```
[...]

Issue #1327, opened 2022-07-07
Airspy in Macos


loewal:
Airspy HF+ and Discovery don't work in Macos, neither selfcompiled nor DMG.
Windows and Linux work fine.

loewal:
May I express my disappointment for not having recieved any reaction on my problem?

The Airspy's are found, but do NOTHING

lawendel:
I'm observing the same (lack of) behavior on a MacBook Air M1 under Monterey. I've downloaded the i64 DMG and installed flawlessly. The program opens with no error messages, it then recognizes the HF+ as an input device, it can even switch between HF and VHF, but nothing happens in the Spectrum window. Before reading your issue, I thought I made something wrong. Maybe it's that... Is there a mailing list or a group where we can ask directly to f4exb? SDRangel is indeed complex but highly attractive as a cross-platform app. I'm also using GQRX and SDR++ with no such issues.

lawendel:
Sorry, it's the recent "macx64" DMG...

justfalter:
I have an Airspy r2 on an M1, compiled for arm64. 

A single breakpoint set in [AirspyWorker::rx_callback](https://github.com/f4exb/sdrangel/blob/d1a4fca49bcabf847f36b595f6b47202895d8080/plugins/samplesource/airspy/airspyworker.cpp#L279-L291) would never hit after starting the input.

After some debugging,  I found I could get the Airspy input working properly by:
1. Setting a breakpoint right before the `m_airspyWorkerThread` is started [here](https://github.com/f4exb/sdrangel/blob/d1a4fca49bcabf847f36b595f6b47202895d8080/plugins/samplesource/airspy/airspyinput.cpp#L240)
2. Starting the input
3. Hit the breakpoint
4. Wait a couple seconds
5. Continue execution

My guess is that there is some kind of timing issue in the worker thread startup sequence.

The effective sequence is:
* `m_airspyWorker->startWork()`
  * `airspy_stop_rx(...)`
  *  (on success): `airspy_is_streaming(...)`
* `m_airspyWorkerThread.start()`

I'm unfamiliar with the mechanics of a QThread, as that is how [m_airspyWorkerThread is defined](https://github.com/f4exb/sdrangel/blob/d1a4fca49bcabf847f36b595f6b47202895d8080/plugins/samplesource/airspy/airspyinput.h#L142).


[...]

PR #1436, opened 2022-09-21
Update threading in airspyhf source

srcejon:
Update airspyhf source to threading model for #1346 which gets it working in MacOS for me, hopefully fixing #1327

I guessed the problem was calling airspyhf_start in a different thread to the callback.


loewal:
Can confirm that it is working for me also Jon!

Great work!!

PR #1437, opened 2022-09-22
Spectrum: Add peak power and frequency in info bar

srcejon:
This patch adds the peak power and frequency into the spectrum info bar.

image

Currently always enabled, but could add a button to make it optional if desired.

### DIFF
--- a/sdrgui/gui/glspectrum.cpp
+++ b/sdrgui/gui/glspectrum.cpp
@@ -593,7 +593,6 @@ void GLSpectrum::newSpectrum(const Real *spectrum, int nbBins, int fftSize)
     QMutexLocker mutexLocker(&m_mutex);
 
     m_displayChanged = true;
-
     if (m_changesPending)
     {
         m_fftSize = fftSize;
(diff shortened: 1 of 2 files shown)


f4exb:
> Currently always enabled, but could add a button to make it optional if desired.

I am not definite on this let's give it a bit of practice

f4exb:
I understand the functionality and motivation but I see a few shortcomings:

- it is not tremendously useful if the signals are transient
- frequency display is down to the Hz regardless of the FFT bin size. I think precision should be down to 10^log10(binSize). For example if bin size is between 10 Hz and 100Hz it should be enough to have a precision down to 10Hz. This would limit distraction with the low digits changing all the time and limit width to only what is useful. Note that this may apply to the center frequency display too and I am not sure it is presently.
- the space occupied is not always void. When markers are displayed and the frequency of the marker is above the high limit of the bandwidth then frequency is displayed in the same space resulting in a garbled display. See screenshot next

image

So at least in a first approach yes this should be made optional.

For the longer run... I already thought of improving the markers so that they are not fixed in frequency but can be made to track the bin with maximum magnitude as one can find on many spectrum analyzers. One can imagine tracking the current max or the max of max depending on the choice of marker type: https://github.com/f4exb/sdrangel/blob/master/sdrgui/gui/spectrummarkers.md#9-marker-type Thus the max of transient signals could be maintained.

Then one could choose which marker data to display on the top right corner thus achieving the same functionality and more and also solving the display mixing. There are many things on the plate right now and for now just making the display optional would be enough.

I will open a specific issue for markers enhancement.

srcejon:
I'll add a patch to make it optional.

PR #1446, opened 2022-09-25
Spectrum measurements

srcejon:
This PR changes the Peak measurement patch, to add several different selectable spectrum measurements, including:

- Peak
- Channel power
- Adjacent channel power
- SNR
- SNFR
- THD
- THD+N
- SINAD
- SFDR
- (and None, for no measurement)

See the files changed for some example images and docs.

Note that I haven't yet changed the precision used for the peak, as I wanted the exact bin frequency, and you can't get this from the markers (it seems the spectrum isn't quite plotted on the bin frequency) - so I'll address that in a later patch. But for now, this at least restores the old appearance.


f4exb:
I think the functionality is interesting but this is not something everyone will use every day so we have to think of something less prominent in the GUI.

srcejon:
How about a simple measurements toggle icon in the controls GUI, that shows/hides a measurements widget that opens to the side as you suggest? 

That would also have the benefit of being able to display SNR/THD/SINAD simultaneously which would be preferable anyway (along with a peak table, and a few other goodies I was thinking of adding later).

f4exb:
This is how it looks like when spectrum display is minimized:
image

I think it needs a bit of refinement...

f4exb:
This could also be a Rx channel on its own as we already have the Channel Analyzer.

f4exb:
We definitely need more space for display and control. See this for source of inspiration: https://www.youtube.com/watch?v=ypu5glggyLI

f4exb:
The resolution bandwidth us already displayed as a tool-tip over the FFT size selection combo box.

srcejon:
> This could also be a Rx channel on its own as we already have the Channel Analyzer.

But I would like the measurements to be available for all spectrum views. They're not all applicable in every case, but channel power can be used for TX, and THD might be used for audio spectra, etc.

f4exb:
OK... that's a valid point and there is a spectrum display in the channel analyzer anyway so this can be used to isolate that spectrum display from the main spectrum avoiding collision with the other spectrum overlays like the channel markers.

srcejon:
Sounds good - I'll do an updated patch.

PR #1479, opened 2022-10-16
Adjust number of frequency digits displayed depending whether tranverter is enabled

srcejon:
So most users who don't use the transverter only seem the relevant number of digits for their device, but allow all to be displayed for those who want them.

Just done for a few devices, to see if the patch will be approved. Will follow up with the rest if it is.


f4exb:
Not sure it is worth the extra code. I don't mind the extra digits after all and like the systematic approach. I will have to think of it.

f4exb:
OK... this makes sense.

PR #1508, opened 2022-11-09
Device GUI window size constraints

srcejon:
I noticed a problem whereby somewhat randomly, a Device GUI would not be displayed properly. One way to repeatedly reproduce this issue is to maximize the window (although there currently isn't any code to do that - I found it while trying to track down the original problem).

This problem appears to be due to the way the size of windows are currently constrained, by having a resizeEvent handler, that ignores the requested size, and calls adjustSize(), which resizes to the minimum size hint instead. I think this code was added to prevent the Tile windows button from making the Device windows too big, as the Tile code seems to ignore the Fixed size policy.

However, it seems the problem can be fixed by calling setMinimumSize() and setMaximumSize() for the window, and Tile respects these constraints.

To avoid hardcoding sizes, I added the sizeToContents() function, which calculates the min/max window size from the .ui file. It also picks up the sizePolicy from the .ui file as well, so the hardcoding for that has been able to be removed.

With the problem resolved, it's now possible to call showMaximized(), without windows being corrupted. I therefore thought it might be useful to add a Maximize window button, as we already have a Shrink button:

image

However, most of the Device GUIs are fixed size, so these buttons and the resize grip are redundant, so I thought I'd hide them in sizeToContents() if the size policy was Fixed in both directions. This helps to declutter the title bar a little bit:

image

If you're happy with the changes, I can then do the same for Channels and Features. (I'm not particularly bothered about the Maximize button, if you don't want it. More useful for some of the Features/Channels though)


f4exb:
> If you're happy with the changes, I can then do the same for Channels and Features.

I have prospected this a little bit but I don't think this will work. The "content" in devices is a simple `QWidget` so its minimum and maximum sizes are set via the `.ui` file. However for Channel and Features this is a `RollupWidget` which is quite different. At least for the height the minimum and maximum is computed by the `arrangeRollups()` method so the `sizeToContents()` method will have no effect. Maximize seems to maximize in width but not in height. For the height it will just leave the current height.

srcejon:
Yeah, it will need to work a little differently, as the min/max size can change when the widget is rolled. So we should be able to have something similar to sizeToContents() in either onWidgetRolled() or perhaps on the LayoutRequest event.

f4exb:
OK, my remark was just out of warning.

PR #1510, opened 2022-11-11
Maximize windows part 2

srcejon:
Add maximize button to MainSpectrum and expandible Channels and Features.

Add sizeToContents in ChannelGUI and FeatureGUI, called when widget is rolled, so we can remove resizing code from all of the individual channels and features.

In RollupContents, use minimumSizeHint for calculated size, so that minimumWidth can come from .ui file.

In DeviceGUI::sizeToContents(), call adjustSize(), so Device GUIs start out at minimum needed size (which should restore appearance prior to last patch).

In stackSubWindows, use available space for channels if no spectrum/features present.
In stackSubWindows, fix spectrum from being sized too big, resulting in scroll bars appearing.    
Reset user-defined channel width in stackSubWindows, when channels are removed.
Don't stack maximized windows.

There's one hack in Channel/FeatureGUI::maximizeWindow(). It seems that when maximimzing a window, QOpenGLWidgets aren't always paint properly immediately afterwards, so the code forces an additional update. I can't see why the first call to paintGL doesn't work.

Let me know if you see anything odd.

### DIFF
--- a/plugins/channelmimo/beamsteeringcwmod/beamsteeringcwmodgui.cpp
+++ b/plugins/channelmimo/beamsteeringcwmod/beamsteeringcwmodgui.cpp
@@ -16,7 +16,6 @@
 ///////////////////////////////////////////////////////////////////////////////////
 
 #include <QLocale>
-#include <QResizeEvent>
 
 #include "device/deviceuiset.h"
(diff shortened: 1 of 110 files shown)


f4exb:
Looks good to me thanks!

PR #1518, opened 2022-11-17
Qt6 support

srcejon:
I saw a blog post recently that suggested Qt Location will soon be available for Qt 6. This is the one missing component of Qt 6 that we need support for.

So I thought I'd take a look at seeing what we needed to change in order to build with Qt 6. Fortunately, not a huge amount. The major change is that different classes are used for audio. The changes for that are in a separate commit. I had to write a wrapper class around the device info (AudioDeviceInfo), as the new class, QAudioDevice, is not compatible with QAudioDeviceInfo. There are still quite a few #ifdefs in the audio code, that could be refactored, if you prefer.

I found and fixed a few small bugs along the way. One bug I haven't fixed, is that it will crash if trying to restore settings that have multiple workspaces or devices. A single device seems to work though, and I've been able to use most of the Demods and Features. (Apart from those that use Qt location for the map). (Multiple devices work when opened in the workspace - just seems to be an issue loading from settings).

The code of course should still compile with Qt5. To use Qt6, you need to run cmake with -DENABLE_QT6=ON.

I've only tested with Qt6 on Windows, as it's easy to do. That was with Qt 6.4.0. 6.4.1 has a bug that means there's no audio output on Windows.

(I'm not suggesting we move to Qt6 or anything - I guess it will take a while to appear in Linux distros - just fancied giving it a try to see what is new. Should be easy enough to support both after this)


f4exb:
For Ubuntu 22.04 it seems I get most dependencies with this:
```
sudo apt-get update && sudo apt-get -y install \
qt6-base-dev \
libqt6websockets6-dev \
qt6-multimedia-dev \
libqt6serialport6-dev \
libqt6core5compat6-dev \
libqt6multimedia6 \
libqt6opengl6-dev \
libqt6quick6 \
libqt6charts6-dev \
qml6-module-qtpositioning \
qml6-module-qtquick-window \ 
qml6-module-qtquick-dialogs \
qml6-module-qtquick-controls \
qml6-module-qtquick-layouts \
libqt6serialport6-dev \
qt6-declarative-dev \
qt6-webengine-dev \
qt6-base-private-dev
```
But this is Qt 6.2.4 and it misses the `TextToSpeech` module that only appears in 6.4 so this cannot work (yet).

srcejon:
Ah, didn't realise 6 was available. I think TextToSpeech can be made optional fairly easily. I'll give it try.

PR #1519, opened 2022-11-17
Enable compilation with Qt 6.2

srcejon:
Make TextToSpeech optional component and add needed header.

Made WebEngine optional, as it's not useful without Location anyway.


f4exb:
In addition to the previous list you will probably need this package as well: qt6-webengine-dev-tools

f4exb:
Provided that last dependency it does compile in Ubuntu 22.04 without the ADSB demod but that's expected.

PR #1541, opened 2022-12-20
Window arrangement

srcejon:
Add buttons to stack MDI windows vertically and put in tabs.
Use right click to auto-stack sub-windows, rather than having a dedicated button.
Allow maximize button to make window full screen, if already maximized.
Add title to device windows, for when displayed in tabs.
Add menu button to workspace toolbar, for Android only, to avoid having menu bar, which takes up a lot of space.
Add configuration presets button to workspace toolbar. (Will need a following patch to work)
Add icons for window arrangement - please feel free to make some better ones!

### DIFF
--- a/sdrbase/settings/configuration.cpp
+++ b/sdrbase/settings/configuration.cpp
@@ -65,6 +65,13 @@ QByteArray Configuration::serialize() const
         s.writeBool(301 + i, m_workspaceAutoStackOptions[i]);
     }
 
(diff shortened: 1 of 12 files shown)


f4exb:
Icons are OK. Maybe the tab arrangement one is a bit cryptic but it is difficult to represent something in such a tiny space. Tooltip is there anyway.

PR #1542, opened 2022-12-20
MainWindow updates for Android and example configurations

srcejon:
Allow sampling devices to be refreshed while SDRangel is running. (Possibly need to add a mutex and MainCore::releasePluginManager to go with MainCore::getPluginManager, so only can be used by one thread)
Add option to automatically update My Position based on GPS. This is started in MainCore, so we get position as soon as possible. 
Don't set QFileDialog::DontUseNativeDialog on Android, as Qt's file dialog can't access user storage. Set globally for Linux/Windows, rather than for each dialog.   
 
MainWindow:
- Add welcome dialog for Android.
- Don't show menu bar or status bar on Android to save screen space.
- On Android, change tab position when screen orientation changes.
- Load example configurations and presets first time SDRangel is run.
- Change loadConfiguration to use a QProgressDialog rather than QMessageBox, as the latter was crashing on Android. (Possibly still needs to run in a thread.)
- Use DialogPositioner to ensure dialogs are on screen.


f4exb:
So is it now possible to generate an .apk for Android? I suppose one needs to set up a special environment for this (cross compiler at least). Are there or will there be instructions to do this?

srcejon:
Need a few more patches. Also need built libraries and I had to build a custom version of Qt in order to fix a couple of bugs. Yep, you will need a cross compiler. Will add some instructions eventually.

PR #1543, opened 2022-12-21
Scan for serial AMBE devices on Windows

srcejon:
Currently the AMBE engine doesn't scan for serial devices on Windows. There the comment:

     // Do not activate serial support at all for windows

Any idea why? Bad idea to send commands to all serial devices without knowing what they are?

This PR adds the scanning, which seems to find the device.

### DIFF
--- a/plugins/feature/ambe/ambeengine.cpp
+++ b/plugins/feature/ambe/ambeengine.cpp
@@ -78,10 +78,27 @@ void AMBEEngine::getComList()
     }
 }
 
-// Do not activate serial support at all for windows
 void AMBEEngine::scan(QList<QString>& ambeDevices)
 {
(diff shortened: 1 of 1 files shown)


f4exb:
> Bad idea to send commands to all serial devices without knowing what they are?

You can't do otherwise. The USB vendor:serial `0403:6015` is just generic FTDI controller. Nothing to tell it is the AMBE dongle.

srcejon:
> You can't do otherwise. The USB vendor:serial `0403:6015` is just generic FTDI controller. Nothing to tell it is the AMBE dongle.

Could just list the available serial ports though, and then let the user select one? QSerialPortInfo::availablePorts(); should list them.

Having said that, it is convenient that it only shows AMBE devices.

PR #1554, opened 2023-01-02
Android related changes

srcejon:
An assortment of changes for Android.

Main one for you to review is probably the changes to the File Input plugin. Android has strange permissions model for accessing user storage, that requires use of URLs. QFile supports this, but istream doesn't, so had to change the I/O functions to use QFile. Currently this is done via #defines, but QFile code could just be used on all platforms. Let me know if you want me to simplify this.


f4exb:
Hello and Happy New Year!

It is fine for me to use QFile in place of istream and simplify the code.

PR #1612, opened 2023-03-03
Disable some warnings when compiling with MSVC

srcejon:
So more important ones aren't hidden in the noise.


srcejon:
This is failing because of the update to the DAB library - which is dealt with in #1610.

Issue #1625, opened 2023-03-08
Compilation failure with Qt 6


Mohamedadlyi:
These lines of codes failed to build with Qt 6.2.4 with Core5Compat
The errors I got are:

(1)
https://github.com/f4exb/sdrangel/blob/54288cf4c70e33e0588b3e3f8e6bb68b2b41de7b/sdrbase/maincore.cpp#L361-L362
```
sdrangel/sdrbase/maincore.cpp:361:60: error: ‘updateTimeout’ is not a member of ‘QGeoPositionInfoSource’
  361 |         connect(m_positionSource, &QGeoPositionInfoSource::updateTimeout, this, &MainCore::positionUpdateTimeout);
      |                                                            ^~~~~~~~~~~~~
sdrangel/sdrbase/maincore.cpp:362:106: error: no match for call to ‘(const QOverload<QGeoPositionInfoSource::Error>) (QGeoPositionInfoSource::Error (QGeoPositionInfoSource::*)() const)’
  362 |         connect(m_positionSource, qOverload<QGeoPositionInfoSource::Error>(&QGeoPositionInfoSource::error), this, &MainCore::positionError);
      |           
```
(2)
https://github.com/f4exb/sdrangel/blob/54288cf4c70e33e0588b3e3f8e6bb68b2b41de7b/sdrgui/mainwindow.cpp#L177-L180
'''
sdrangel/sdrgui/mainwindow.cpp:177:15: error: ‘class QScreen’ has no member named ‘setOrientationUpdateMask’
  177 |     screen()->setOrientationUpdateMask(Qt::PortraitOrientation
      |               ^~~~~~~~~~~~~~~~~~~~~~~~
'''
(3)
https://github.com/f4exb/sdrangel/blob/54288cf4c70e33e0588b3e3f8e6bb68b2b41de7b/sdrgui/gui/dialogpositioner.cpp#L32-L35

'''
sdrangel/sdrgui/gui/dialogpositioner.cpp:32:23: error: ‘class QScreen’ has no member named ‘setOrientationUpdateMask’
   32 |     dialog->screen()->setOrientationUpdateMask(Qt::PortraitOrientation
'''
(4)
https://github.com/f4exb/sdrangel/blob/54288cf4c70e33e0588b3e3f8e6bb68b2b41de7b/plugins/channelrx/demodft8/ft8demodsettingsdialog.h#L24
'''
(5)
[...]

PR #1629, opened 2023-03-15
fixing errors when compiling using qt6

Mohamedadlyi:
#1625 fixing errors when compiling using qt6

### DIFF
--- a/sdrbase/maincore.cpp
+++ b/sdrbase/maincore.cpp
@@ -358,7 +358,9 @@ void MainCore::initPosition()
     if (m_positionSource)
     {
         connect(m_positionSource, &QGeoPositionInfoSource::positionUpdated, this, &MainCore::positionUpdated);
+        #if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
         connect(m_positionSource, &QGeoPositionInfoSource::updateTimeout, this, &MainCore::positionUpdateTimeout);
+        #endif
         connect(m_positionSource, qOverload<QGeoPositionInfoSource::Error>(&QGeoPositionInfoSource::error), this, &MainCore::positionError);
         m_position = m_positionSource->lastKnownPosition();
         m_positionSource->setUpdateInterval(1000);
@@ -398,6 +400,12 @@ void MainCore::positionUpdateTimeout()
 
 void MainCore::positionError(QGeoPositionInfoSource::Error positioningError)
 {
+    #if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    if (positioningError == 4){
+        positionUpdateTimeout();
+        return;
+    }
+    #endif
     qWarning() << "MainCore::positionError: " << positioningError;
 }
 
--- a/sdrgui/gui/dialogpositioner.cpp
+++ b/sdrgui/gui/dialogpositioner.cpp
@@ -29,10 +29,12 @@ DialogPositioner::DialogPositioner(QWidget *dialog, bool center) :
     m_center(center)
 {
     connect(dialog->screen(), &QScreen::orientationChanged, this, &DialogPositioner::orientationChanged);
+    #if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
     dialog->screen()->setOrientationUpdateMask(Qt::PortraitOrientation
                                                 | Qt::LandscapeOrientation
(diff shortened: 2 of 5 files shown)


srcejon:
Rather than add a new macro (DENABLE_QT6), Qt defines version numbers that can be used:

#if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))

srcejon:
This is failing on Qt 5, as size() returns an int.

    averagePeriod->setMinimum(std::max((long long){1}, m_averagePeriodTexts.size() - value)

Whereas Qt 6 size() returns qsizetype (which is ssize_t). Should be safe in this instance to cast size() to an int, as the list is always going to be small.

sdrbase/maincore.cpp
```diff
@@ -358,7 +358,9 @@ void MainCore::initPosition()
     if (m_positionSource)
     {
         connect(m_positionSource, &QGeoPositionInfoSource::positionUpdated, this, &MainCore::positionUpdated);
+        #if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
         connect(m_positionSource, &QGeoPositionInfoSource::updateTimeout, this, &MainCore::positionUpdateTimeout);
+        #endif
```
f4exb, line 363: no #else ?

sdrbase/maincore.cpp
```diff
@@ -398,6 +400,12 @@ void MainCore::positionUpdateTimeout()
 
 void MainCore::positionError(QGeoPositionInfoSource::Error positioningError)
 {
+    #if (QT_VERSION >= QT_VERSION_CHECK(6, 0, 0))
+    if (positioningError == 4){
+        positionUpdateTimeout();
+        return;
+    }
+    #endif
```
f4exb, line 408: The return in the $if block is not very good practice

sdrgui/gui/dialogpositioner.cpp
```diff
@@ -29,10 +29,12 @@ DialogPositioner::DialogPositioner(QWidget *dialog, bool center) :
     m_center(center)
 {
     connect(dialog->screen(), &QScreen::orientationChanged, this, &DialogPositioner::orientationChanged);
+    #if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
     dialog->screen()->setOrientationUpdateMask(Qt::PortraitOrientation
                                                 | Qt::LandscapeOrientation
                                                 | Qt::InvertedPortraitOrientation
                                                 | Qt::InvertedLandscapeOrientation);
+    #endif
```
f4exb, line 37: no #else ?

sdrgui/mainwindow.cpp
```diff
@@ -174,10 +174,12 @@ MainWindow::MainWindow(qtwebapp::LoggerWithFile *logger, const MainParser& parse
 	connect(&m_inputMessageQueue, SIGNAL(messageEnqueued()), this, SLOT(handleMessages()), Qt::QueuedConnection);
 
     connect(screen(), &QScreen::orientationChanged, this, &MainWindow::orientationChanged);
+    #if (QT_VERSION < QT_VERSION_CHECK(6, 0, 0))
     screen()->setOrientationUpdateMask(Qt::PortraitOrientation
                                       | Qt::LandscapeOrientation
                                       | Qt::InvertedPortraitOrientation
                                       | Qt::InvertedLandscapeOrientation);
+    #endif
```
f4exb, line 182: no #else ?

Mohamedadlyi: No else because the function setOrientationUpdateMask is removed  sice qt 6 and the property comes by default set to all these four orientations so no need for calling it ,same apply to another comment of yours

sdrbase/maincore.cpp
Mohamedadlyi: Since qt6 the signal updateTimeout is removed and got added to the errors enums as an erorr ,so no need for this connection,instead i made if condition in your error slot function  to call the slot function of timeout error

sdrgui/gui/dialogpositioner.cpp
f4exb: OK as per next comment

sdrbase/maincore.cpp
f4exb: OK as per previous comment

Issue #1657, opened 2023-04-14
Support Adalm Pluto on Mac


loewal:
Sorry if this is not the right place:

I would like support fot the Adalm Pluto in the MacOs DMG-version.

Thanks

srcejon:
Currently ENABLE_IIO=OFF for the DMG build, because current build of libxml2 requires autoconf which is a dependency I didn't want on Mac.

However, it looks like it may be able to be built with cmake instead - so I'll give that a try.

loewal:
Thanks in advance Jon


Regards

srcejon:
Give this a try - No idea if it works, as I don't have a pluto.

https://sdrangel.org/downloads/sdrangel-7.13.0_10.14.6_x86_64.dmg

loewal:
Yes Jon. It works!
Ever so grateful. Will support be permanent now?

Kind regards

srcejon:
Implemented.

PR #1663, opened 2023-04-18
Update libxml2 for libiio

srcejon:
Updates libxml2 used by libiio, to the latest version, so it can be built using cmake and there's no autoconf dependency. This allows Pluto support in Mac .dmg for #1657.


loewal:
👍🍀

Issue #1605, opened 2023-02-23
MacOSX M1/M2 support


oxoocoffee:
Hello

I am just a new user of SDRAngel. Just downloaded the latest "heat map" version and realized it is only x64 version. After installing it should work through an emulator. But when I try to start it I get this message "You can’t open the application “SDRangel” because it may be damaged or incomplete."

I just cloned source and will take a look at it. A quick cmake run produces error

```
CMake Error at cmake/Modules/DetectArchitecture.cmake:193 (message):
  Flag '-march=native' rejected by compiler.  Please adjust ARCH_OPT option.
Call Stack (most recent call first):
  cmake/Modules/CompilerOptions.cmake:3 (include)
  CMakeLists.txt:570 (include)
```

I will take a stab at DetectArchitecture.cmake and add a fix. Any other setup/build  pointers welcome.

73,

satdna:
I have the same problem with compilation on Apple SoC (M2). I will appreciate any help.

oxoocoffee:
I will try this weekend if I have some time. Else weekend after I will have time to try

oxoocoffee:
Ok did make a bit of progress. I have Qt6 installed. Also, all additional packages are installed using [homebrew](https://brew.sh).

```
brew install boost         # 1.81.0_1
brew install opencv      # 4.7.0
brew install codec2      # 1.0.5
brew install hidapi         # 0.13.1
```

I am not worrying about SDRplay for now. Will look at this later.
What is the list of any additional packages that need to be installed and supported version?

I can run cmake with the following arguments
```
cmake  -DENABLE_SDRPLAY=OFF -DCMAKE_MACOSX_RPATH=1 -DENABLE_QT6=1 ..
```

It did produce this output. Can any developer help me out if the above arguments are correct? I like to get it close as possible to actual release builds.

```
-- The C compiler identification is AppleClang 14.0.0.14000029
-- The CXX compiler identification is AppleClang 14.0.0.14000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
[...]

PR #1668, opened 2023-04-24
Rotator controller: Add gamepad/joystick control

srcejon:
Add support for gamepad/joystick control of the Rotator Controller's target and offset.

This uses Qt Gamepad, and so there are two new package dependencies, libqt5gamepad5-dev and libqt5gamepad5.

I have tried making this optional in CMakeLists.txt, however, it seems on Ubuntu, find_package(qt ... OPTIONAL_COMPONENTS Gamepad) reports an error if the gamepad package isn't installed, rather than just setting the FOUND variable to false.


srcejon:
Not sure where the libqt5gamepad5-dev dependency for ubuntu2204docker needs to be added. (Have done ubuntu2204prod)

f4exb:
> edit: sdrangel-docker/sdrangel/Dockerfile maybe?

Yes I think so.

vielmetti:
Not finding this library in Homebrew, resulting in an error in the M1/M2 build logged at https://github.com/f4exb/sdrangel/issues/1605#issuecomment-1606012236

PR #1693, opened 2023-05-15
DSC demod

srcejon:
Hi Edouard. This PR adds a DSC (Digital Selective Calling) demod used by marine radios.

Also makes a few minor fixes to some other plugins.


srcejon:
And have also added a slot map to the AIS Demodulator.

Issue #1682, opened 2023-05-02
10kHz doppler errors when simultaneous pass


Valstee:
Hi @f4exb and @srcejon , 
I use sdrangel (SDRangel  7.8.3-8-g0c9686511  Qt 5.15.3  x86_64  Ubuntu 22.04.2 LTS) to receive amateur radio satellites like uvsqsat, inspiresat-7 and many others.
I don't know how to explain it but lately I have been getting regular 10kHz doppler errors when uvsqsat and inspiresat-7 are passing by simultaneously.
The prioritization works well, i.e. it targets the right satellite but the doppler starts at +/- 20kHz and goes to about 0Hz which is fine but 10kHz off... 
I have made presets for each of the satellites and when they don't arrive at the same time I don't have this problem. 
Do you know where this can come from?
Thanks,
F4JIZ

Valstee:
yes sorry, I waited to have again the same configuration (i.e. a simultaneous passage of the 2 satellites).

```
2023-05-17 11:24:43.755 (D) SatelliteTrackerWorker::doppler  "UVSQ-SAT"
2023-05-17 11:24:43.756 (D) SSBDemod::webapiSettingsPutPatch: forward to GUI: 0x55c975d36638
2023-05-17 11:24:43.756 (D) SatelliteTrackerWorker::doppler  "INSPIRE-SAT 7"
2023-05-17 11:24:43.756 (D) SSBDemod::webapiSettingsPutPatch: forward to GUI: 0x55c975d36638
2023-05-17 11:24:43.760 (D) SSBDemod::handleMessage: MsgConfigureSSBDemod
[...]

PR #1696, opened 2023-05-23
Fix Sat Tracker crash

srcejon:
Just a bit too late... Fix Sat Tracker crash introduced by last patch for #1682.

Add support for tracking Satellites in the Star Tracker, to enable satellites to be used as calibration signals in Radio Astronomy plugin.


f4exb:
np that's what minor releases are there for

srcejon:
Thanks!

Issue #1740, opened 2023-07-15
DAB demodulator plugin missing in Linux deb package for version 7.15.1


computer-wizard:
The dab demodulator plugin is missing in the Linux deb package for version 7.15.1.
I checked the /usr/lib/sdrangel/plugins sub-directory and the libdemod library for 
dab isn't there.

The only library related to dab is /usr/lib/sdrangel/libdab_lib.so but as I already said earlier,  
there is no libdemod library for it :-(

srcejon:
@f4exb - How can I see the generated CMakeCache.txt for the ubuntu2204prod build on appveyor to debug? Does it need to be added as an additional artifact in .appveyor.yml, or is is possible to ssh in?

In the ubuntu2204prod  log file, it seems the dab project is built, so I presume both ZLIB_FOUND AND FAAD_FOUND must be true, and that should set DAB_FOUND to TRUE.

The demoddab build is conditional on:

    if(ENABLE_CHANNELRX_DEMODDAB AND DAB_FOUND AND ZLIB_FOUND AND FAAD_FOUND)
        add_subdirectory(demoddab)
    endif()

So can't then see why it wouldn't then be built.

srcejon:
It looks like when AUTO_EXTERNAL_LIBRARIES is set in external/CMakeLists.txt, the FOUND variables aren't cached, so they are undefined in the plugins CMakeLists.txt

From the new log:

Not building demoddab (ENABLE_CHANNELRX_DEMODDAB=ON DAB_FOUND=ON ZLIB_FOUND= FAAD_FOUND=)
[00:23:06] -- Not building sigmffilesink (ENABLE_CHANNELRX_SIGMFFILESINK=ON LIBSIGMF_FOUND=)
[00:23:06] -- Not building modatv (ENABLE_CHANNELTX_MODATV=ON OpenCV_FOUND=)

computer-wizard:
Good find. I wonder why they aren't cached?

f4exb:
Appveyor also has its own build cache. I just cleaned it.

srcejon:
> Good find. I wonder why they aren't cached?

I think it's a scope problem.

When AUTO_EXTERNAL_LIBRARIES is set, the find_package variables from external/CMakeLists.txt are lost at the end of that script.

Recent versions of cmake have a global flag to find_package, that would keep them in scope, but the more compatible solution, I think, is just to move the find_package commands in to the root CMakeLists.txt. So, I'll give that a try...

[...]

PR #1747, opened 2023-07-29
Display message if channelrx plugins not built.

srcejon:
For #1740 - but also useful generally.

Also add missing enable flag for heatmap.


srcejon:
The prod build just seemed to hang after 7 mins:

    [00:07:11] 
    [00:07:11] 80% [1155 libqt5location5 1,165 B/756 kB 0%]                                                                                                                                                                                     
    3,275 kB/s 50s

f4exb:
If you give me your username on Appveyor I can give you build permission which should do the trick. However the Appveyoor builds run on my own server so  usually when a build fails I do some cleanup and restart the Appveyor agent. However it may work.

PR #1750, opened 2023-07-30
Move find_package in external/CMakeLists.txt to CMakeLists.txt

srcejon:
Move find_package in external/CMakeLists.txt to CMakeLists.txt to see if variables remain in scope. (Needs to be tidied up if it works)
Also fix codec2 build.
Add status messages for features build.


srcejon:
Still have one more problem - The libsigmf install is being included in the .deb file, which we don't want.

srcejon:
To clarify the problem:

Most of the projects in external/CMakeLists.txt have the INSTALL_COMMAND set to "" and the libs and includes are picked up from the build directory rather than the install directory (which doesn't exist).

However, libsigmf seems to create some needed files only during the install, so we need to run the install step for that. 

The deb packager is picking up those installed files, which we don't want (apart from the .so, which needs to be moved to /usr/lib/sdrangel instead).

Perhaps this is why we have the cmake/ci/build_*.sh scripts that build some of the libraries that install in to /usr/local/ instead.

Haven't used the deb packager before, so working my way though the process...

PR #1765, opened 2023-08-06
Fix/typos

dforsi:
This PR fixes a typo in the name of a private method and the TLDs in two context menus.


dforsi:
The list of beacons at https://www.ncdxf.org/beacon/ is not the same as the one in `ibpbeacon.cpp`:

- KH6RS vs KH6WO
- VR2B vs VR2HK

srcejon:
Yep, looks like they are out of date.

PR #1766, opened 2023-08-08
Add support for JupiterSDR

catkira:
https://wiki.analog.com/resources/eval/user-guides/jupiter-sdr

### DIFF
--- a/devices/CMakeLists.txt
+++ b/devices/CMakeLists.txt
@@ -26,6 +26,7 @@ endif()
 
 if(ENABLE_IIO AND LIBIIO_FOUND)
     add_subdirectory(plutosdr)
+    add_subdirectory(jupitersdr)
 endif()
 
 if(ENABLE_PERSEUS AND LIBPERSEUS_FOUND)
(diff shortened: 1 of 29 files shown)


catkira:
@f4exb It still takes some more commits to get this working. I just created this PR already so that You can give early feedback.

f4exb:
Looks like a nice SDR. Since it is a transceiver and has two synchronized inputs and outputs it may be better supported as a "mimo" device. I see it uses libiio and you probably got inspiration from PlutoSDR. Input and Output device plugin for the same device is the way of support prior to the "mimo" devices implementation. PlutoSDR has the 3 flavors of support: input, output and mimo.

catkira:
ah thanks, I will have a look at the pluto MIMO device.

catkira:
the inactivity is unfortunately caused by ADI, because with the ADRV its not possible anymore to change the sample rate by just writing an iio attribute. With these devices one has to upload a binary blob (so called profile) that contains configured settings such as sample rate. ADI said they want to open source the code to generate the binary blob, but so far it has not been done.

f4exb:
The "feature" label should take this PR out of the no activity scope

f4exb:
I had a quick look at the documentation and what they call "profiles". This looks like part of the device firmware from the software point of view so there is no point in trying top change these parameters from software. However it is necessary to be able to read essential parameters like the sample rate. I did not go that far to know if this is possible but I would hope so.

Issue #1166, opened 2022-02-25
CUDA Accelerated FFT


f4exb:
**Is your feature request related to a problem? Please describe.**
Large FFT with large overlap is too slow

**Describe the solution you'd like**
I would like the FFT to run on my NVIDIA GPU and just for the sake of it!

**Describe alternatives you've considered**
FFTW sucks!

**Additional context**
For wider GPU support this could be considered: https://github.com/DTolm/VkFFT

Jcwscience:
I have tried a GitHub project called radio-core which uses cuda acceleration for Broadcast FM demodulation, it seems to make a significant impact in reducing system load.

Jcwscience:
Also I’m using a hackrf so there is a lot of data being processed, from what I understand the hackrf internal filters only work properly with a sample rate of over 10mhz (although I might have misunderstood)

Jcwscience:
But if I’m being honest my main motivation probably is “well the sdr software _I use_ has **gpu acceleration**, look at my cool setup with all of this compute power!”.

f4exb:
That's what I meant: GPU acceleration just for the sake of it....

Jcwscience:
@f4exb actually I was just looking at vkfft, I hadn’t seen it before, if it doesn’t have any breaking bugs then maybe it could be useful for people with amd cards as well? Also I hope I’m not asking too many questions but other than the waterfall is there anything else that uses FFTW? I’m fairly new to this field entirely but I’m enjoying studying the code to see how things work.

srcejon:
There are a couple of other plugins that use FFTs as well, but all the code does this indirectly via the FFTEngine and FFTFactory classes. 

So what you probably want to look at, is modifying the FFTFactory to return a VkFFTEngine (which would be a subclass of FFTEngine) if it is applicable for the current system - if not, return an FFTWEngine - or something along those lines.

See sdrbase/dsp/fftwengine.h kissengine.h (which is an alternative to FFTW) and FFTFactory.cpp - It looks like it should be fairly straightforward to drop a different implementation in there.

While I doubt it will be of much benefit to existing plugins - high performance FFT and IFFT could be useful for OFDM modems in the future.

Jcwscience:
@srcejon Awesome, I saw reference to an alternative fft engine in the source files, but I didn’t quite know where it fit into the rest of things. I went ahead and forked the repo and I’m running some benchmarks on VkFFT now as well.

f4exb:
[...]

PR #1779, opened 2023-08-14
GPU FFT and simple profiler

srcejon:
This PR adds support for running FFT on GPU using the VkFFT library, with either Vulkan or CUDA backend, as per #1166

To allow easier performance comparisons, I have updated the FFTEngine and FFTFactory class to support multiple simultaneous FFTEngines, so a user can switch between FFTW, Kiss and VkFFT at run-time via a GUI option (Preferences > FFT). Also, different bits of SDRangel can request different FFTEngines. So FFTW could be used generally, but a specific plugin could request VkFFT if it thinks it will benefit from it.

I have also added a simple profiler, that displays average execution time in a dialog (View > Profile Data...). Profiling is enabled via the cmake -DENABLE_PROFILER=ON option. It is disabled by default, and when disabled, there should be no performance impact as the profiling code is not included in the executable. It can also be used to easily profile other parts of SDRangel - here showing GL spectrum draw time, as well as FFT. To profille some code, you just need to stick the PROFILE_START() and PROFILE_STOP("name") macros around the code.

image

Whether VkFFT uses Vulkan or CUDA is a cmake option, -DVKFFT_BACKEND=0 for Vulkan and -DVKFFT_BACKEND=1 for CUDA. The code should compile without VkFFT if neither the Vulkan or CUDA SDKs are available.

On Windows, VkFFT is automatically downloaded as an external project. If external project is off, then it can be downloaded with:

    cd /opt/build
    git clone https://github.com/DTolm/VkFFT.git
    cd VkFFT
    git checkout v1.3.1

It's a header only library, so doesn't need to be built.

Either the Vulkan or CUDA SDKs need to be installed to compile it:

Linux Vulkan Setup

    cd /opt/install
    mkdir -p vulkansdk
    cd vulkansdk
    wget https://vulkan.lunarg.com/sdk/download/1.3.250.1/linux/vulkansdk-linux-x86_64-1.3.250.1.tar.gz
    tar xf vulkansdk-linux-x86_64-1.3.250.1.tar.gz
    source /opt/install/vulkansdk/1.3.250.1/setup-env.sh
    ...
    cmake -DVKFFT_INCLUDE_DIR=/opt/build/VkFFT/vkFFT/ -DVKFFT_BACKEND=0 ..

Linux CUDA setup

    Install CUDA toolkit from https://developer.nvidia.com/cuda-downloads
    export PATH=/usr/local/cuda/bin:%PATH
    ...
    cmake -DVKFFT_INCLUDE_DIR=/opt/build/VkFFT/vkFFT/ -DVKFFT_BACKEND=1 ..

Windows Vulkan Setup

    Install Vulkan SDK (With optional 64-bit Debug libraries) from https://vulkan.lunarg.com/sdk/home#windows
    Copy include\glslang\Include\glslang_c_interface. and include\glslang\Include\glslang_c_shader_types.h from Linux SDK as they are missing (See https://vulkan.lunarg.com/issue/view/63d158a85df11200d569b2ab)
    set VULKAN_SDK=C:\VulkanSDK\1.3.250.1  (or where ever you installed it)
    cmake -DVKFFT_BACKEND=0 ..

Windows CUDA Setup

    Install CUDA toolkit from https://developer.nvidia.com/cuda-downloads (This creates CUDA_PATH environment variable)
    cmake -DVKFFT_BACKEND=1 ..

Unfortunately, on my PCs, using the GPU (via either Vulkan or CUDA) is much slower than using FFTW, due to the overhead of getting the data from the CPU to the GPU. It may be more beneficial on embedded SoCs that have unified memory and relatively slower CPUs, but haven't tried. To get better performance, we'd probably need to submit FFTs in batches, rather than one at a time (or do additional processing on the GPU). So there's no reason to try to enable this in the release builds now, but it's the first step towards GPU support.


f4exb:
> To get better performance, we'd probably need to submit FFTs in batches, rather than one at a time (or do additional processing on the GPU)

The problem in this case is that the rendering will probably be sketchy. It may be more beneficial to processes using FFT heavily in the background.

RobRich999:
Just rolled an updated local build. I will probably look into profiling actual performance later, but for now according to mesa debugging info, VkFFT under SDRangel appears to be working okay on my Kubuntu 23.04 system with AMDPU and RADV.

AMD 5700u processor here, so AMD Vega GPU with 8 CUs.

Thanks to the project devs for the feature and support. :)

savagesmc:
Thank you for adding this feature!

Issue #1430, opened 2022-09-20
User-defined default settings / settings templates for features and channels using preset concept


srcejon:
As previously discussed: https://github.com/f4exb/sdrangel/issues/1139#issuecomment-1049661748, I think it would be useful if users could set the initial settings for any particular plugin, so that whenever that feature or channel is added, the user's settings are automatically applied.

This would be particularly useful for features such as the Map, that have settings such as API keys, that you don't want to have to re-enter each time you open the map. But it can be useful in channels too - E.g. you might always want the Broadcast FM demod to start in stereo with RDS enabled.

While this is somewhat supported via presets, loading a preset changes all of the channels / features, which isn't always what is wanted. E.g. if you add the Map into an existing setup, you can't load a preset with the Map, as that might close other features.

Could perhaps be implemented by adding save & reset buttons in the Common Settings dialog.

f4exb:
> I think it would be useful if users could set the initial settings for any particular plugin, so that whenever that feature or channel is added, the user's settings are automatically applied.

Well... maybe it is not the easiest nor the safest way to implement this. I'd rather have it as presets but unlike the current presets they would be of a kind that applies only to the channel or feature type. At the expense of having to do things in two steps (1: add the channel/feature with default settings, 2: select and load the preset) it has the following advantages:

  - we do not change the way a channel or feature is added. This is a sensitive part of the code and it is safer not to fiddle with it too much
  - you can have more than one preset for the channel or feature type. For example with SSB demod you can have several filter settings for wider or narrower SSB, choice between LSB/USB or CW
  - You are not tied to the same settings to create every channel or feature
  - similarly to what exists today for presets you would also be able to add, change, export and import such presets
  - presets are a known and old story it would be not much different in functionality than the present presets only the scope changes. Maybe some parts of the code can be reused or at least be a source of inspiration.

Note that if you want to add many channels of the same type you can create the first and then duplicate it. I don't think this exists for features. It is probably less useful but could be implemented the same way.

Note also that it means changing a bit the way current feature presets are implemented as they load a collection of features. There would be a bit of reorganization to support both single and multiple feature presets.

srcejon:
Having multiple per-individual channel / feature presets sounds good.

But I would really like it, if one could load automatically :) 

[...]

PR #1789, opened 2023-08-24
Add support for Plugin Presets

srcejon:
This PR adds support for Plugin Presets (or Device/Channel/Feature Presets), as discussed a while back in #1430.

This allows Presets to be saved / loaded for individual Devices, Channels and Features, via the new Star button in the basic settings dialog. E.g:

image

image

It should basically work the same as the existing Device Set and Feature Set presets, except that the preset only applies to a single Device, Channel or Feature.

image

If a preset is saved with Group/Description as Defaults/Default, then it will be loaded automatically whenever a plugin of that particular type is created.

### DIFF
--- a/sdrbase/CMakeLists.txt
+++ b/sdrbase/CMakeLists.txt
@@ -207,6 +207,7 @@ set(sdrbase_SOURCES
 
     settings/configuration.cpp
     settings/featuresetpreset.cpp
+    settings/pluginpreset.cpp
     settings/preferences.cpp
     settings/preset.cpp
     settings/mainsettings.cpp
@@ -437,6 +438,7 @@ set(sdrbase_HEADERS
 
     settings/configuration.h
     settings/featuresetpreset.h
+    settings/pluginpreset.h
     settings/preferences.h
     settings/preset.h
     settings/mainsettings.h
--- a/sdrbase/device/deviceapi.cpp
+++ b/sdrbase/device/deviceapi.cpp
@@ -482,10 +482,10 @@ QList<quint64> DeviceAPI::getCenterFrequency() const
     }
     else if (m_deviceMIMOEngine && m_deviceMIMOEngine->getMIMO())
     {
-        for (int i = 0; i < m_deviceMIMOEngine->getMIMO()->getNbSourceStreams(); i++) {
+        for (uint32_t i = 0; i < m_deviceMIMOEngine->getMIMO()->getNbSourceStreams(); i++) {
             frequencies.append(m_deviceMIMOEngine->getMIMO()->getSourceCenterFrequency(i));
         }
-        for (int i = 0; i < m_deviceMIMOEngine->getMIMO()->getNbSinkStreams(); i++) {
+        for (uint32_t i = 0; i < m_deviceMIMOEngine->getMIMO()->getNbSinkStreams(); i++) {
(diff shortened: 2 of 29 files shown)


srcejon:
It appears Appveyor may increase build time limit to 90 minutes if we ask.

Maybe we can speed it up as well. It seems a bit variable .

srcejon:
Ah, I just remember you said this is on your own server. I could probably set up one as well, if it helps.

srcejon:
I think there are probably still a couple of things to do / might need adjusting:

- I haven't made any changes to the API to support this.
- As is, settings are serialized via the GUI classes. E.g. ChannelGUI rather than ChannelAPI. Perhaps it should have been the latter, so the presets can be loaded in to GUIless version? For channels, they both seem to serialize the same objects, so may work anyway, but for Devices, it actually serializes the DeviceUISet to include MainSpectrum settings along with the device settings (but not it's channels).

f4exb:
> As is, settings are serialized via the GUI classes. E.g. ChannelGUI rather than ChannelAPI. Perhaps it should have been the latter..

Yes, this is preferable. 

> ... but for Devices, it actually serializes the DeviceUISet to include MainSpectrum settings...

You can look at what is done for channels that have a spectrum display like the SSB demod. A pointer to the spectrum settings is stored in the settings as a "Serializable" object. It may be present or not (nullptr) for GUIless version. Serializable is a pure virtual class (interface) exposing only pure virtual methods like "serialize" or "deserialize". Then the demod settings object uses these methods to serialize or deserialize to/from a blob.

srcejon:
> > As is, settings are serialized via the GUI classes. E.g. ChannelGUI rather than ChannelAPI. Perhaps it should have been the latter..
> 
> Yes, this is preferable.

Is the GUI object not potentially a superset of settings? E.g. it should contain all of the settings of the underlying object, plus potentially additional GUI specific settings?

> > ... but for Devices, it actually serializes the DeviceUISet to include MainSpectrum settings...
> 
> You can look at what is done for channels that have a spectrum display like the SSB demod. A pointer to the spectrum settings is stored in the settings as a "Serializable" object. It may be present or not (nullptr) for GUIless version. Serializable is a pure virtual class (interface) exposing only pure virtual methods like "serialize" or "deserialize". Then the demod settings object uses these methods to serialize or deserialize to/from a blob.

Yes, that's what I'm doing - but via DeviceUISet, rather than DeviceAPI, as DeviceAPI doesn't have a pointer to the Spectrum - it's only in the DeviceSet/DeviceUISet.

f4exb:
The GUI object and the object it represents have each their own copy of the settings. Both copies get synchronized in one way or the other depending if the origin of the change is the GUI or the represented object via API.

srcejon:
> Both copies get synchronized in one way or the other depending if the origin of the change is the GUI or the represented object via API.

Unfortunately, it seems like this isn't currently quite the case just using serialize/deserialize. We'd have to use the Web API.

If I call XXXDemodGUI::deserialize() (as the current patch does), then the settings are indeed passed to the ChannelAPI object (via XXXDemodGUI::applySettings() then a MsgConfigureXXX).

However, if I try to do the other way, by calling XXXDemod::deserialize() or XXXFeature::deserialize(), settings aren't propagated back to the GUI (settings are applied by sending a MsgConfigureXXX to the ChannelAPI object)

It seems settings are only propagated to the GUI, when webapiSettingsPutPatch is called.

PR #1790, opened 2023-08-25
Fix RTLSDR E4000 gain and bandwidth settings and add RTLSDRBlog V4 support

srcejon:
I noticed two problems with Nooelec RTLSDR with E4000 tuner:

- Gain setting was not working. After adjusting the gain, it would re-adjust itself as though AGC had been reenabled. It appears we need to call rtlsdr_set_tuner_gain_mode before each call to rtlsdr_set_tuner_gain.
- Adjusting RFBW would result in a bizarre frequency response.

Also, added a change to the GUI that displays the tuner type and disables Ofs checkbox if not supported for the given tuner.

### DIFF
--- a/external/CMakeLists.txt
+++ b/external/CMakeLists.txt
@@ -22,7 +22,7 @@ set(MIRISDR_TAG "v2.0.0")
 set(LIBSIGMF_TAG "new-namespaces")
 
 set(PTHREADS4W_TAG "Version-2-11-0-release")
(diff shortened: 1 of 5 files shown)


srcejon:
Also added support for RTLSDRBlog V4, by updating to latest RTLSDR driver and setting low high frequency limit to 0kHz to support it's integrated HF upsampler.

PR #1809, opened 2023-09-05
Command line options to start Remote TCP Sink

srcejon:
The following patch adds command line options to be able to start the Remote TCP Sink on a specified device, so you can start a remote server without using the API or GUI, which should hopefully make it easier to use. E.g:

    sdrangelsrv --remote-tcp --remote-tcp-hwtype RTLSDR

Additional options:

    --remote-tcp-serial: Specify device by serial number, instead of a hwtype
    --remote-tcp-address: TCP address of interface to listen on
    --remote-tcp-port: TCP port to listen on
    --list-devices: List physical devices, so you can see what needs to be passed to --remote-tcp-hwtype/serial

All the device settings (sample rate / freq etc) can be set by the client Remote TCP Device GUI, so shouldn't need command line options.

Thoughts?

### DIFF
--- a/app/main.cpp
+++ b/app/main.cpp
@@ -36,6 +36,7 @@
 
 #include "loggerwithfile.h"
 #include "mainwindow.h"
+#include "remotetcpsinkstarter.h"
 #include "dsp/dsptypes.h"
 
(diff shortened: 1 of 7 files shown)


f4exb:
looks good to me

Issue #1839, opened 2023-09-22
NSIS issue Win 11 


m6mal:
Tried to install on my 17-1265 laptop and my i7-12700 pc and it throws up "installer integrity check has failed, common causes include incomplete download and damaged media. Contact the installer's author to obtain anew copy"  and will not install at all this is reffering to the version i cant install is V7.16.0 and is only 57Mb and the last version 7.15.4 is 146MB

f4exb:
The file size starts to drop signifcantly from there: https://github.com/f4exb/sdrangel/actions/runs/6196072083

m6mal:
So why cant mine install  yesterdays version v7.16 as i get the error NSIS issue

m6mal:
just see you have reupped it in downloads and still getting instal error nsis

PR #1841, opened 2023-09-24
Github Actions updates

srcejon:
Hi Edouard,

This PR does two main things:

- Remove building of .zip on Windows in order to free up disk space, and allow the .exe to be built for #1839. (I think the .zip isn't used anyway).
- Add a build on MacOS using QT6. This doesn't produce a releasable .dmg yet, but seems a good way to check that the code compiles without error on the Mac and also with Qt 6, helping catch errors before I go to do a release build.

Note that I've changed the way the version number is obtained for the .exe, as the existing method didn't work when the Action was run in a forked repository. It's now read from the version number used for CPack that I've written to CMakeCache.txt, which should be more reliable that using the git tag. 

Added a command in the Windows build to dump out amount of disk space after the build. After removing the .zip build, we now have 0.67 GB left (Github provides 14GB on D drive).

It also fixes a couple of problems in the source, shown when building on the Mac using Qt6 using the new action.

Yeah - it took a few goes!


f4exb:
Seems to fix the Windows build issue. Well done!

PR #1854, opened 2023-10-22
Fix 2 race conditions

srcejon:
This PR fixes 2 race conditions in some RX demods.

The first is the initialisation of the number of streams used by a scope. Currently this is performed in the GUI, but this can be too slow, if the sink gets samples before the GUI initialisation code has run, resulting in a crash. So I've moved the init code to the baseband, where the scope is created. 

The next problem appears if a channel is added to a device set that is already running. When this happens, the baseband class initially receives a DSP signal notification with a baseband sample rate of 0. This results in the DownChannelizer not being initialised properly. For some demods with low sample rates, it seems Baseband::handleData() can be called before the correct baseband sample rate is received and used to correctly initialise the DownChannelizer. When this happens, the channels will hang (somewhere in m_channelizer->feed), the log will show samples being dropped and if you try to stop the device, SDRangel will hang completely.

This patch has a little workaround for the demods I've noticed have this problem (on my computer, at least). In the handleData() method, I've added a check to make sure a valid baseband sample rate has been set in the channelizer, preventing m_channelizer->feed from being called, if not. E.g:

Old:
    while ((m_sampleFifo.fill() > 0) && (m_inputMessageQueue.size() == 0))
New:
    while ((m_sampleFifo.fill() > 0) && (m_inputMessageQueue.size() == 0) && m_channelizer->getBasebandSampleRate())

Is there a better way? (Perhaps DownConverter could have a isValid() method, to also check channelSampleRate is correct etc)

Should this code be added to all demod basebands? I guess the problem potentially applies everywhere.

### DIFF
--- a/plugins/channelrx/demodais/aisdemodbaseband.cpp
+++ b/plugins/channelrx/demodais/aisdemodbaseband.cpp
@@ -32,6 +32,7 @@ AISDemodBaseband::AISDemodBaseband(AISDemod *aisDemod) :
 {
     qDebug("AISDemodBaseband::AISDemodBaseband");
 
+    m_scopeSink.setNbStreams(AISDemodSettings::m_scopeStreams);
     m_sink.setScopeSink(&m_scopeSink);
     m_sampleFifo.setSize(SampleSinkFifo::getSizePolicy(48000));
     m_channelizer = new DownChannelizer(&m_sink);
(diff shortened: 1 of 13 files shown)


srs4511351:
I am having an issue where the scope in the DSD Demodulator will freeze.
I don't know if it is related to this issue.

It will work initially, but after a period of time, it will freeze. The length of time varies.
Closing DSD Demodulator and reopening it does not restore the scope functionality. It is necessary to quit SDRangel and restart it. This has been occurring for many months.

I am running a Raspberry Pi 4 with Raspberry Pi OS Bookworm.
SDRangel 7.17.0

----Steve

PR #1855, opened 2023-10-23
Add separate audio bandwidth setting to AM Demod

srcejon:
This patch adds a setting to AM Demod to allow audio bandwidth to be set independently from RF bandwidth.

This is particularly useful when listening to ATC that uses offset carrier / CLIMAX, which requires much wider RF bandwidth than audio bandwidth. E.g. spectrum looks like this:

image


srcejon:
Unlike with SSB, with AM the carrier offset doesn't affect the demodulated audio frequency, so you hear normal voice.

You set RF bandwidth wide (~20kHz) in order to be able to receive transmissions with no offset (aircraft) and an offset (airport ground stations). After demod, you get the audio signal from both centered at 0Hz, and then you want to filter the audio above 5kHz, because the that's how wide the audio signal is.

The reason this is done, I believe, is to support what seems to be a form of antenna diversity. i.e. two signals are transmitted (with the same audio, but from different locations), one at a positive offset and one at a negative offset.

Issue #1836, opened 2023-09-21
DSD Demodulator scope display freezes


nostar:
The DSD scope freezes after doing things like stopping and starting the device or changing device frequency.  It stays frozen untill sdrangel is restarted.  This is git source build on Fedora using either rtl-sdr or HackRF.

srcejon:
Yep - can see this happening on Windows too.

simonmarty:
Was able to reproduce on macOS and release v7.16.0.

srcejon:
One thing to check is the squelch setting. 

But it seems it can stop even when that is at the minimum setting

c3kkos:
I have an issue with DSD demodulator. 

I can open another issue here but i think its more polite just to add content.

so here i am, i setup a basic usb rtl sdr and trying to decode something from the air.

the samples are recieved but dsd is dead. what am i doing wrong?

https://github.com/f4exb/sdrangel/assets/30070329/5bf173e9-1cf2-4798-9176-c7e627d15480

PR #1883, opened 2023-11-09
DSD Demod - Fix loss of constellation

srcejon:
DSD Demod constellation will not appear if !m_running when GUI is created or if acquisition is restarted (which recreates the baseband object).

This patch fixes this, by saving a pointer to m_scopeXYSink in DSDDemod, so that it can be used to initialise the baseband/sink every time it is created.

Also, this patch adds a warning in the DSDDemod GUI if the sample rate is < 48kHz.

For #1836


srcejon:
Also added another patch, so that the AudioSelectDialog uses the DialogPositioner so that it is fully visible on screen.

Issue #1826, opened 2023-09-12
M17 packet mode needs an update


sp5wwp:
We have recently updated the M17 protocol's specification document. SDRangel has issues with decoding packets. Could you apply a few minor adjustments to the code to accommodate for that?

The SMS should have the following structure (packet data contents):
0x05 (text message designator), text message (UTF-8 encoded), 0x00 (null-termination), 2-byte CRC (same as for the rest of the protocol).
The message is segmented into 25-byte chunks. If there's not enough data left, the rest is zero-padded. We kept the bit order for the Packet Metadata Field you used before - EOT bit is `1<<7` and the data byte counter goes to `<<2`. I believe the CRC has wrong endianness though. Most significant byte should be at position `[0]` of the byte array.

sp5wwp:
It is possible that only the CRC endianness bug is preventing the messages to be displayed.

sp5wwp:
Since SDRangel is the most convenient way to send M17 packets, could we have this update?

srcejon:
SDRangel m17 modem is based on this https://github.com/mobilinkd/m17-cxx-demod/commits/master/include/m17cxx - which doesn't seem to have had any changes since it was forked.

In M17Modulator.h make_lsf, there is:

        std::array<uint8_t, 2> checksum = crc.get_bytes();
        lsf[28] = checksum[0];
        lsf[29] = checksum[1];

And make_packet_frame:

            packet_assembly[packet_size]   = crc_.get_bytes()[1];
            packet_assembly[packet_size+1] = crc_.get_bytes()[0];

Which looks the opposite way around. If we reverse it, in m17demodprocessor.cpp, we'd then also need to reverse:

       uint16_t calcChecksum = crc16.get_bytes()[0] + (crc16.get_bytes()[1]<<8);

srcejon:
I don't have a radio to test with m17-tools - but the above patch should reverse the CRC byte ordering in packet mode, so hopefully you can test it? I just made sure it still worked SDRangel to SDRangel.

srcejon:
Oh, just re-read your second comment, and it seems there may be another change needed. Can you clarify what you think is wrong? (I don't actually use it...)

sp5wwp:
You can TX M17 using rpitx if you have a Raspberry Pi around. We need to compare what SDRangel expects in a packet (what is extracted from it per frame) and what the reference code actually puts there.

srcejon:
[...]

PR #1891, opened 2023-11-14
Reverse byte ordering of CRC in packet mode in M17 modem

srcejon:
For #1826, reverse byte ordering of CRC in packet mode in M17 modem


srcejon:
I wouldn't commit this yet - looks like some other changes are needed.

f4exb:
I was too quick...

PR #1892, opened 2023-11-15
M17: Update packet format to comply with current spec

srcejon:
As discussed in #1826

### DIFF
--- a/modemm17/M17Modulator.h
+++ b/modemm17/M17Modulator.h
@@ -128,16 +128,12 @@ public:
         std::copy(source_.begin(), source_.end(), rit);
 
         lsf[12] = can_ >> 1;
-        lsf[13] = (streamElsePacket ? 5 : 4) | ((can_ & 1) << 7);
+        lsf[13] = (streamElsePacket ? 5 : 2) | ((can_ & 1) << 7);
 
         if (gnss_on_)
         {
             lsf[13] |= (1<<5);
             std::copy(gnss_.begin(), gnss_.end(), &lsf[14]);
-        }
-        else
-        {
-            lsf[13] |= (3<<5);
         }
 
         crc.reset();
--- a/plugins/channelrx/demodm17/m17demodprocessor.cpp
+++ b/plugins/channelrx/demodm17/m17demodprocessor.cpp
@@ -251,10 +251,10 @@ void M17DemodProcessor::decode_type(uint16_t type)
                 m_typeInfo += "UNK";
                 break;
             case 1:
-                m_typeInfo += "RAW";
+                m_typeInfo += "DAT";
                 break;
             case 2:
-                m_typeInfo += "ENC";
+                m_typeInfo += "UNK";
                 break;
             case 3:
                 m_typeInfo += "UNK";
--- a/plugins/channeltx/modm17/m17modprocessor.cpp
+++ b/plugins/channeltx/modm17/m17modprocessor.cpp
@@ -60,6 +60,7 @@ bool M17ModProcessor::handleMessage(const Message& cmd)
         const MsgSendSMS& notif = (const MsgSendSMS&) cmd;
         QByteArray packetBytes = notif.getSMSText().toUtf8();
         packetBytes.prepend(0x05); // SMS standard type
+        packetBytes.append('\0'); // SMS should be null terminated
         packetBytes.truncate(798); // Maximum packet size is 798 payload + 2 bytes CRC = 800 bytes (32*25)
         processPacket(notif.getSourceCall(), notif.getDestCall(), notif.getCAN(), packetBytes);
         // test(notif.getSourceCall(), notif.getDestCall());
--- a/sdrgui/gui/mdiutils.cpp
+++ b/sdrgui/gui/mdiutils.cpp
@@ -18,6 +18,7 @@
 #include <QWidget>
 #include <QDebug>
 #include <QDataStream>
+#include <QIODevice>
 
 #include "mdiutils.h"
 


sp5wwp:
Have you brought back the CRC to its original byte order?

srcejon:
No - I think I was looking at the wrong diff when I said it doesn't need to be changed, as I was working on 2 different computers.

With this patch, you should get data that matches what was discussed in #1826. If you don't let me know and I'll take another look.

sp5wwp:
I'm unable to setup my PlutoSDR to transmit valid M17 packets. For some reason the sample rate of the transmitted signal is much too high.
obraz
obraz

srcejon:
Why do you have BB set to that rather unusual value of 65104?

How are you determining that the sample rate is to fast?

sp5wwp:
65.104k happens to be the lowest value I can get. I think it is too fast because I can see a large part of the transmission in the SDR++'s demodulator window.

f4exb:
65.104k * 32 = 2083.33k (appears in the upper right) which is effectively the lowest sample rate you can obtain with the Pluto without using the FIR decimating filter (FIR button + decimation combo next). However I am not sure sample rate is the issue because the modulator should accommodate the required sample rate to the baseband sample rate with the rational interpolator. I suppose it needs at least 48k so anything higher than this is fine.

sp5wwp:
We need someone else than me to check this out, as I'm clueless what's going on.

srcejon:
> 65.104k happens to be the lowest value I can get. I think it is too fast because I can see a large part of the transmission in the SDR++'s demodulator window.

Just gave it a quick try here, SDRangel to SDRangel - and it doesn't appear to work in packet mode, if baseband sample rate is < 100k.

However, even using a sample rate > 100k, the m17 demod in SDR++ doesn't receive the packet.

If I look at a received signal using SDRangel's channel analyzer, the preamble is 40ms:

image

EoT may be a little short:

image

But the symbol rate looks OK. 8*1/4800=1.66ms

image

sp5wwp:
I have just recorded M17 stream sample, as obtained with the newest SDRangel. Looks like the preamble is ~550Hz tone, 170ms long, instead of 40ms of 2400Hz. Math looks correct too: 2400*(40/170)=565.
obraz

srcejon:
Odd - because I don't see that. Can I suggest you try running your Pluto at a sample rate of 1MS/s.

sp5wwp:
2,083,336Sa/s is the lowest I can go.

srcejon:
Ok, use that, and I will too.

sp5wwp:
BTW - `RFBW` and `Dev` sliders should be removed, but that's another story :)

srcejon:
Ok, I see something odd. When I used the SRC callsign of SP5WWP, SDRangel stops displaying the packets.

Can I suggest you duplicate the setup shown here:

image

As that's the packet we confirmed with the reference encoder, and I can see working here.

You can measure the packet transmission time, by pressing the pause button highlighted in red:

image

Then press the marker button and select Show All then Close.

The in the waterfall, hold shift, and left click at the start and end of the packet. It should presumably be 160ms as in the pic above.

sp5wwp:
Both B210 and PlutoSDR transmit too long frames.

obraz

srcejon:
> Both B210 and PlutoSDR transmit too long frames and nothing is being decoded.

I've no idea how that's possible, without me seeing the same.

srcejon:
> Ok, I see something odd. When I used the SRC callsign of SP5WWP, SDRangel stops displaying the packets (it's reporting a CRC error).

This looks like a bug in M17DemodProcessor::decode_lsf()

    if (!lsf[111]) // LSF type bit 0
    {
        uint8_t packet_type = (lsf[109] << 1) | lsf[110];

It's using bit numbers as indexes in to a byte array. Can be fixed with:

    if (!(lsf[13] & 1)) // LSF type bit 0
    {
        uint8_t packet_type = (lsf[13] >> 1) & 0x3;

sp5wwp:
Was `decode_lsf()` originally located in the library you use? Is the bug coming from the lib itself?

srcejon:
> Was `decode_lsf()` originally located in the library you use? Is the bug coming from the lib itself?

Doesn't look like it.

srcejon:
> Both B210 and PlutoSDR transmit too long frames and nothing is being decoded.

Can you measure the time in the TX spectrum, rather than the RX spectrum., please? All I can think of, is for some reason (E..g PC/USB not fast enough), the samples aren't being output at the correct rate.

sp5wwp:
> > Both B210 and PlutoSDR transmit too long frames and nothing is being decoded.
> 
> Can you measure the time in the TX spectrum, rather than the RX spectrum., please? All I can think of, is for some reason (E..g PC/USB not fast enough), the samples aren't being output at the correct rate, unless there's an uninitialised variable somewhere..

Same issue. TX time is longer than vertical range.

obraz

srcejon:
If you increase the FFT size, you should be able to increase the vertical range.

sp5wwp:
obraz

srcejon:
> > BTW - `RFBW` and `Dev` sliders should be removed, but that's another story :)
> 
> I can only receive audio stream in SDR++ if I increase the SDRangel Dev setting above 5.7k (default is 5k). The closest match to the SDR++ reference lines is at 8k.

So, from the M17 specification, FM peak deviation should be 2.4k, but SDRangel currently has a default setting of 5k.

However, setting the symbol output to constant 1, a setting of 5k doesn't actually result in a deviation of 5kHz.

The problem is in M17ModSource::modulateSample(). Currently, we have:

    m_modPhasor += (m_settings.m_fmDeviation / (float) m_audioSampleRate) * t1;

Where m_fmDeviation is 10,000 if GUI setting is 5k. (i.e. full deviation)

However, I think we actually want 2 * pi * peak deviation, so:

    m_modPhasor += ((2.0f * M_PI * (m_settings.m_fmDeviation / 2.0f)) / (float) m_audioSampleRate) * t1;

And with that, I then see a deviation of 2.4kHz with GUI setting of 2.4kHz - and audio is correctly received in SDR++.

On the SDRangel receive side, I think a factor of PI is missing in the demod too:

        m_phaseDiscri.setFMScaling(48000.0f / (2.0f*settings.m_fmDeviation));

Should be:

        m_phaseDiscri.setFMScaling(48000.0f / (2.0f*M_PI*settings.m_fmDeviation));

sp5wwp:
>  FM peak deviation should be 2.4k, but SDRangel currently has a default setting of 5k

Actually, the peak instantaneous deviation is somewhat larger than 2.4kHz, as the baseband is RRC filtered and exhibits a small overshoot. But yes, of course +3 symbol should give +2.4k deviation. Otherwise, your maths makes much sense, well spotted. Can the sliders be removed to avoid confusion and misuse? I think we are going to fix a few bugs along the way, hehe :)

srcejon:
> Actually, the peak instantaneous deviation is somewhat larger than 2.4kHz, as the baseband is RRC filtered and exhibits a small overshoot. But yes, of course +3 symbol should give +2.4k deviation. 

Yep.

This is the t1 signal in the above equation:

image

I checked a value of 1 results in a deviation of 2.4kHz, and >1 will give that overshoot.

PR #1894, opened 2023-11-17
Fix M17 packet type decoding.

srcejon:
Fix M17 packet decoding - bit indices were being used in to a byte array, as mentioned in #1892.

Also remove copy+paste error in GUI tooltip.


srcejon:
Also fix same bug that was in DSD Demod that was causing constellation to sometimes not be displayed.

srcejon:
Add patch so that FM frequency deviation is +/- 2.4kHz.

PR #1897, opened 2023-11-18
Add FRS-GMRS presets

sudo-n00b:
Hey, thank you devs for the awesome work. I use SDRAngel to see all the radios being used around me, this preset would help.


sudo-n00b:
I have another preset for FRS+GMRS+MURS+CCR
I also have .csv file for annotation markers for all these freqs if it is helpful for anyone

Issue #1901, opened 2023-11-23
Channel analyzer: channel scope trigger


keukenrol:
Hi

I am trying to get the triggers working in the channel scope, but it does not seem to function as expected.
I have set trace length to a minimum to detect a single pulse. With a sample rate of 20MSPS, this gives me a trace length of 240 microseconds.
I would like to use the one shot trigger. The level is known, thus it is set using the Trig0 trigger level slider. Tried both real, imaginary and magnitude variants. While the signal does show up in freerunning mode, it never seems to actually trigger.

Am I missing something?

Thanks!

srcejon:
Can you add a screenshot please?

keukenrol:
> Can you add a screenshot please?

I suppose it would be easier if I make a video of it, but here is a screenshot with a summary:
![Trig_F](https://github.com/f4exb/sdrangel/assets/56192644/92402f3f-6f75-4522-8295-c0cfe2174afe)
In freerunning mode (thus, trigger not active), it is clear that the signal level rises above the set trigger level. The green trigger line is below the pulse amplitude. The trigger level is set very low, just above the noise level to make sure it triggers as soon as possible.

![Trig_O](https://github.com/f4exb/sdrangel/assets/56192644/ea0dcdd4-2716-46f2-b9bf-ce655a2567a2)
In oneshot, the image remains frozen like this. I would expect to be able to trigger on a pulse as soon as it rises above the set trigger level, and freeze the display.

I also noticed that the framerate of the channel scope is only a few FPS. While the channel spectrum easily runs on 50FPS.

srcejon:
For the waveform I'm looking at, it triggers in one-shot mode, but not where I'd expect it to:

![image](https://github.com/f4exb/sdrangel/assets/57259258/b261a3d7-f6f1-46a1-bd37-d91dfac9510a)

With rising edge trigger, I'd expect it to trigger where the signal crosses upwards across the trigger level, and for that to be time 0.

A smaller issue I see, is that when adjusting the trigger level (when not in single-shot mode), the trigger will occur for each adjustment, even if the adjustment to the level remains outside of the range of the signal.

keukenrol:
Do you have the possibility to try on a high sample rate? I have tried it with the RTL-SDR and the HackRF (shown in screenshots).

srcejon:
With the above patch, triggering seems to be improved.

[...]

PR #1902, opened 2023-11-24
Attempt to fix scope trigger

srcejon:
This is a possible fix for #1901, but please review

Currently, the code in processTrace only ever appears to look at the first sample in a block of data

    const Complex& s = *vbegin[triggerStreamIndex];

I'm guessing it needs to be indexed using processed (as that variable isn't read anywhere else in the loop)

    const Complex& s = vbegin[triggerStreamIndex][processed];

And it seems to trigger more as I'd expect.


f4exb:
This seems to make sense

Issue #1933, opened 2023-12-19
ADALM Pluto support in MacOS DMG


tomtastic:
This doesn't appear to be enabled currently in the latest MacOS builds (7.17.2) ?

I would like to see all of this in the MacOS builds if possible? :
* `libiio` enabled
* ADALM Pluto support enabled
* The `plutosdrmimo` plugin enabled

srcejon:
Last time I tried, including libiio was problematic.

tomtastic:
So I had a look at building manually on MacOS (Ventura 13.6.1), as it seems that as `libiio` should build by default, so the ADALM Pluto should be supported without doing anything special.

Whilst doing that I discovered :
### Issue 1 - LAME
Of the external libraries that get pulled down during the build, `LAME` was failing to build it's frontend binaries (which we dont need) due to:
```C
parse.c:417:34: error: call to undeclared function 'id3tag_set_comment_ucs2'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```
So I made a small change to the configure parameters to disable building the LAME frontend binaries :
https://github.com/f4exb/sdrangel/blob/bb406ca9a8188bc814342bbebc1ac38fe7fa65af/external/CMakeLists.txt#L486

I've created a PR for this `LAME` frontend build bug, though it's perhaps unrelated to why `libiio` doesn't appear in the current MacOS releases?

### Issue 2 - UHD
Then I hit a problem building UHD (for USRP?) due to:
```C
/tmp/sdrangel-20231220-29072-lc7i06/sdrangel-7.17.2/build/external/uhd/src/uhd/host/lib/usrp/mpmd/sim_find.cpp:13:10: fatal error: 'Python.h' file not found
```
I've no idea why this happens, as we're passing in `-DPYTHON_EXECUTABLE=/opt/homebrew/bin/python3` to cmake, and I can see the compilation added the appropriate include path `-I/opt/homebrew/Frameworks/Python.framework/Versions/3.11/include/python3.11`, which I confirmed contains the expected `Python.h`.
To workaround this, I added `-DENABLE_USRP=OFF`.
Is this related to UHD missing as reported here https://github.com/f4exb/sdrangel/issues/1870 ?

### Results
Though I can now successfully build (around 1h10mins on my M1 MBP), I am struggling to get cmake to set my the rpath appropriately for the installed location so it can find the shared libraries it's built.
I think mostly a success though?

srcejon:
[...]

PR #1935, opened 2023-12-22
Disable build of LAME frontend

tomtastic:
The external library LAME (v3.100) frontend binaries fail to build (https://github.com/f4exb/sdrangel/issues/1933) on MacOS currently with:
```log
parse.c:417:34: error: call to undeclared function 'id3tag_set_comment_ucs2'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```
We dont need the LAME frontend binaries themselves, only the shared library, so we can disable their build and avoid this bug in their current code.


srcejon:
Haven't seen this problem in my Mac builds, but not building stuff we don't need seems like a good idea to speed up the build anyway.

bardware:
I'm not related to this repository, but I was just compiling lame, got this error and was looking for a solution, how to fix it.
I'm on Windows using MSYS2 environment using the /clang64 tools

Issue #1951, opened 2024-01-13
SDRAngel crashes on PATCH /sdrangel/deviceset/{deviceSetIndex}/channel/{channelIndex}/settings with FreqScannerSettings.


zuker:
STR:
1. Create deviceset with any device, add Frequency Scanner with UI
2. Run

```bash
curl -X PATCH "http://localhost:8091/sdrangel/deviceset/0/channel/0/settings" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{  \"FreqScannerSettings\": {    \"channelBandwidth\": 25000,    \"channelFrequencyOffset\": 25000,    \"channelMarker\": {      \"centerFrequency\": 0,      \"color\": -16724536,      \"frequencyScaleDisplayType\": 0,      \"title\": \"Frequency Scanner\"    },    \"frequencies\": [      {        \"enabled\": 1,        \"frequency\": 100000000      }    ],    \"reverseAPIAddress\": \"127.0.0.1\",    \"reverseAPIChannelIndex\": 0,    \"reverseAPIDeviceIndex\": 0,    \"reverseAPIPort\": 8888,    \"rgbColor\": -16724536,    \"rollupState\": {      \"childrenStates\": [        {          \"isHidden\": 0,          \"objectName\": \"settingsContainer\"        }      ],      \"version\": 0    },    \"streamIndex\": 0,    \"threshold\": -60,    \"title\": \"Frequency Scanner\",    \"useReverseAPI\": 0  },  \"channelType\": \"FreqScanner\",  \"direction\": 0}"
```
ER: Frequemcy Scanner configured
AR: app crashes

PR #1952, opened 2024-01-14
fix SIGSEGV on /sdrangel/deviceset/{deviceSetIndex}/channel/{channelIndex}/settings endpoint

anton-kotenko:
Editing FrequencyScanner via api endpoint cases application crash to access by null pointer.  

Failure happens when swagger generated setValue method is called for complex/nested structure on the object with is created purely by constructor, but not initialised by the ::init method.

Solution:
1. Call init on newly created object
2. Proactively fix this for all types of settings objects

This change addresses https://github.com/f4exb/sdrangel/issues/1951

### DIFF
--- a/sdrbase/webapi/webapirequestmapper.cpp
+++ b/sdrbase/webapi/webapirequestmapper.cpp
@@ -4414,61 +4414,73 @@ bool WebAPIRequestMapper::getChannelSettings(
         if (channelSettingsKey == "ADSBDemodSettings")
         {
             channelSettings->setAdsbDemodSettings(new SWGSDRangel::SWGADSBDemodSettings());
+            channelSettings->getAdsbDemodSettings()->init();
             channelSettings->getAdsbDemodSettings()->fromJsonObject(settingsJsonObject);
         }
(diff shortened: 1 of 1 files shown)


sdrbase/webapi/webapirequestmapper.cpp
```diff
@@ -4480,116 +4492,139 @@ bool WebAPIRequestMapper::getChannelSettings(
...
         else if (channelSettingsKey == "FreeDVModSettings")
         {
             channelSettings->setFreeDvModSettings(new SWGSDRangel::SWGFreeDVModSettings());
+            channelSettings->getFreeDvModSettings()->init();
             channelSettings->getFreeDvModSettings()->fromJsonObject(settingsJsonObject);
         }
         else if (channelSettingsKey == "FreqScannerSettings")
         {
             channelSettings->setFreqScannerSettings(new SWGSDRangel::SWGFreqScannerSettings());
+	        channelSettings->getFreqScannerSettings()->init();
```
anton-kotenko, line 4555: Fix the problem itself. 
Rest of changes are proactively doing the same for the rest of types, to avoid somebody getting into the same trouble

PR #1954, opened 2024-01-17
Sat Tracker: Update default TLEs to latest URLs. 

srcejon:
Update default TLEs to latest URLs as previous GOES TLE no longer exists.
Add GUI error if TLEs not downloaded.


dforsi:
I still get " "Error transferring https://celestrak.org/NORAD/elements/goes.t" if I load the configuration for APT or for ISS Packet, and then I see the old list of TLE sources in Satellite Tracker Settings, but if I click the button that I added in #2151, then I see the new TLE sources that you added in #1957.

Is the list of TLEs also saved in the .prex or .cfgx files? I tried to look into them but I didn't find it.

srcejon:
Yes, the TLEs is saved along with all the other settings. So you need to update manually for any existing configuration / feature set.

PR #1967, opened 2024-01-30
Speedup compilation on all platforms using ccache

GermanAizek:
More information about package: https://ccache.dev


srcejon:
What speed ups do you see?

GermanAizek:
> What speed ups do you see?

Yep, more effective recompilation after checkout branches. It doesn't apply to CI, it's convenience for developers.

PR #1983, opened 2024-02-12
SimplePTT: bugfix and enhancement

mxi-box:
- Fix misuse of device index
- Add "None" option to SimplePTT to disable partial Rx/Tx switching
- Support MIMO device

### DIFF
--- a/plugins/feature/simpleptt/simplepttgui.cpp
+++ b/plugins/feature/simpleptt/simplepttgui.cpp
@@ -288,6 +288,8 @@ void SimplePTTGUI::updateDeviceSetLists()
 
     ui->rxDevice->clear();
     ui->txDevice->clear();
+    ui->rxDevice->addItem(tr("None"), -1);
+    ui->txDevice->addItem(tr("None"), -1);
     unsigned int deviceIndex = 0;
     unsigned int rxIndex = 0;
     unsigned int txIndex = 0;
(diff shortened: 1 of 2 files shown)


plugins/feature/simpleptt/simplepttgui.cpp
```diff
@@ -307,18 +310,23 @@ void SimplePTTGUI::updateDeviceSetLists()
             ui->txDevice->addItem(QString("T%1").arg(deviceIndex), deviceIndex);
             txIndex++;
         }
+        else if (deviceMIMOEngine)
+        {
+            QString text = QString("M%1").arg(deviceIndex);
+            ui->rxDevice->addItem(text, deviceIndex);
```
f4exb, line 316: For MIMO devices you also have to specify a stream index as it handles several I/Q streams simultaneously

mxi-box: > For MIMO devices you also have to specify a stream index as it handles several I/Q streams simultaneously

In SimplePTT's case, it needs only a switch of Rx/Tx.
Seems that specifying stream indexes is unnecessary here, as the MIMO interface is defined without a stream-specified switch.
https://github.com/f4exb/sdrangel/blob/f35ea35a2dae318c6bef63ef99267accc6548c0d/sdrbase/dsp/devicesamplemimo.h#L60

f4exb: I have to check. It means all Rx or Tx streams start and stop at once or just the ones with index 0. If this is not a bug it is a limitation but maybe not to be addressed here.

mxi-box: As far as I see, the current startRx/startTx implementation is switching all streams at once. Maybe an extension is needed for multi-stream devices.

https://github.com/f4exb/sdrangel/blob/f35ea35a2dae318c6bef63ef99267accc6548c0d/sdrbase/dsp/dspdevicemimoengine.cpp#L713

f4exb: In fact this is normal as the different streams are supposed to run synchronously.

PR #1984, opened 2024-02-12
Optimize eumerator of origin device

mxi-box:
- Rewrite **_enumerateDevices_**: support to enumerate all devices in one call
- Add **_enumerateAllDevices_** to reduce startup period
- Optimize LimeSDR enumerating by skipping LMS_Init

i have tested it on LimeSDR for all sinks, sources and MIMOs.
it's my first complicated PR, so please feel free to advise or edit.


f4exb:
Seems to me it is working fine

srcejon:
Looks like it saves ~6 seconds on start-up time, for the set of SDRs I currently have connected. Thanks!

Next trick might be to try to see if the devices can be detected in parallel :)

f4exb:
the parallel thing could be interesting for devices that are looked up on
the network. In particular the Pluto for which I had to remove auto
detection because of too long wait time scanning the network for it.

Issue #1866, opened 2023-10-28
End of Train (EOT, 457.9375 MHz) packets decoder


abelouso:
Hello,
 Would it be possible to add the End of Train packets decoder on 457.9375 MHz? It is a relatively simple known frequency modulated format.

Thank you for reading.

abelouso:
Reference implementation can be found in:

https://github.com/russinnes/EOTDecode/

https://github.com/ereuter/PyEOT

srcejon:
1200k FSK.. can probably use the Packet demod.

Can you capture some IQ recordings?

abelouso:
Recorded I/Q samples (raw) at 457.937500 MHz, 240000 sample rate with decimation of 2: 120.0 ksps
[gqrx_20231231_191106_457937500_120000_fc.raw.gz](https://github.com/f4exb/sdrangel/files/13802052/gqrx_20231231_191106_457937500_120000_fc.raw.gz)

srcejon:
Thanks. Was that train stationary?

If you can get another IQ recording of a train braking, it would be useful, thanks.

f4exb:
DEployed in v7.19.0

abelouso:
Not sure if this could be re-opened or a new enhancement request should be created. 
In the HOT/EOT table there should be an option to update row for an already existing IDs rather than always append received packets as new rows. That would allow to monitor trains by sorting the table columns based on ID, and/or time. Also, adding a signal strength for each received packet as the column in the table would help with approximating the train's proximity and/or direction of travel (away or towards).

Issue #1975, opened 2024-02-06
Link between AIS demodulator and raw signal ?


theoallouche:
Hi,

I have an .sdrqiq file and I'm using the AIS demodulator on it.

I'd like to be able to link the detected messages to the associated IQs, or at least have somehow a timestamp that could link the messages to the raw signal.

How could I do that?
Thanks

srcejon:
At the moment, the AIS timestamp is only the computers time, so you'd have to adjust the timestamps by hand from when you press start. You could possibly do this by importing the csv file in to Excel.

(The satellite tracker can get it's time from the the IQ file, but I think that's the only plugin that does currently).

srcejon:
Have just added a button that allows the choice between having packet date & times come from system clock or File Input device.

PR #2005, opened 2024-03-03
Add End-of-Train demodulator

srcejon:
For #1866

This plugin can be used to demodulate End-of-Train packets. These are packets transmitted by an [End-of-Train Device](https://en.wikipedia.org/wiki/End-of-train_device), that can be found on some trains. It transmits information about whether motion is detected, brake pressue, whether the marker light is on and battery information.


srcejon:
Also, for #1975, I've added a button to AIS, Packet, Radiosonde, DSC and End-of-Train demods, that allows packet date & time to come from File Input (if used) or system clock.

rohrsh:
Heya, can anybody in Australia confirm the ARTC trains do still use these signals?

PR #2027, opened 2024-03-18
Add Airplanes.Live at request of user

adsb-related-code:
Update by request of Reddit user for Airplanes.Live support and port.


adsb-related-code:
Sorry if I didn't do this correctly.

f4exb:
This is just a change in the readme. What's the point?

adsb-related-code:
A user asked this since ADSBExchange was sold to a PE firm in Janaury 2023.  They wanted to know why we weren't listed since we are the tracking site that spawned from the sale and comprised of ex-contributors to that ex-opensource project.  

User is using Windows.  

I thought I would add to readme since other sites are listed too.

PR #2038, opened 2024-03-30
Audio optimize & default input/output bugfix

mxi-box:
Makes every **AudioDeviceInfo** singleton to reduce QAudioDeviceInfo calls. frequently calling WASAPI enumerate when startup will lead to crash on my PC... not really know why but it's an optimization anyway.

Fix things about `m_defaultInputStarted `and `m_defaultOutputStarted`. It creates default devices multiple times, causing totally broken and noncontinuous audio.

### DIFF
--- a/plugins/samplesource/audioinput/audioinputgui.cpp
+++ b/plugins/samplesource/audioinput/audioinputgui.cpp
@@ -186,8 +186,7 @@ void AudioInputGui::updateSampleRateAndFrequency()
 void AudioInputGui::refreshDeviceList()
 {
     ui->device->blockSignals(true);
-    AudioDeviceManager *audioDeviceManager = DSPEngine::instance()->getAudioDeviceManager();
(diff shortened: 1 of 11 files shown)


srcejon:
Is this going to prevent new audio devices from being discovered after the initial enumeration?

mxi-box:
> Is this going to prevent new audio devices from being discovered after the initial enumeration?

As far as Im concerned, there's no way to refresh the device list before. The _AudioSelectDialog_ uses the cached list of AudioDeviceManager, which is not refreshable.
Implementing this feature also involve reindex audio device IDs

srcejon:
> As far as Im concerned, there's no way to refresh the device list before. 

Yeah. 

The problem with this, is the audio devices can change at runtime. For example, on Windows when you connect via RDP. It would be useful to be able to refresh the audio devices so audio can be redirected to the client.

PR #2057, opened 2024-04-06
Support VOR Localizer on Qt6. Add Add Channels dialog to SID. Add JJY to Radio Clock

srcejon:
This PR:

- Updates the VOR Localizer map so that it works on Qt 6.
- Adds a Add Channels dialog in the SID Feature, to easily add Channel Power channels for each VLF signal


srcejon:
Additional patch to add JJY support to Radio Clock. (JJY is Japan's time signal).

PR #2068, opened 2024-04-14
Feature/map feature

dforsi:
This PR:
- avoids unneeded work by updating the IPB Beacons table only when it is visible
- avoids losing dialogs behind the main window adding a parent for:
  - the Beacons dialog
  - the IPB Beacons dialog
  - the Radio Time Transmitters dialog


srcejon:
> * avoids losing dialogs behind the main window adding a parent for:

They're intentionally not parented.

dforsi:
> > * avoids losing dialogs behind the main window adding a parent for:
> 
> They're intentionally not parented.

which I think it's a bad design because AFAICT those are the only windows that are aoutside of the window of sdrangel, but can live with it. However when one of those dialogs is behind the main window, clicking again on its button has no visible effect, maybe it could bring the corresponding dialog to the front?

srcejon:
> which I think it's a bad design because AFAICT those are the only windows that are aoutside of the window of sdrangel, but can live with it. 

If you want to make them proper MDI windows, that's OK, but they shouldn't be stuck in front.

>, maybe it could bring the corresponding dialog to the front?

Yeah, that would make sense.

dforsi:
If I close the main window of sdrangel (with or without the changes in this PR) the 3 dialogs remain open and in this situation when double clicking on a callsign crashes the application, if I don't double click, the application terminates correctly after closing all 3 windows.

Is is intended that thew dialogs do not close when the main window is closed?
Should they be hidden when the map is hidden? How can I close them? The parenting would do that automatically.

srcejon:
They should close when the application closes or the Map feature is closed. This can be fixed by adding the following to the Dialog constructors after ui->setupUi(this);

    setAttribute(Qt::WA_QuitOnClose, false);

We don't want to use parenting as that makes the windows stay on top.

srcejon:
Thanks

f4exb:
@srcejon is it OK to merge now?

srcejon:
Yep, thanks.

PR #2120, opened 2024-05-19
Fix/sid memleaks

dforsi:
This PR deletes some objects and removes some `disconnect()` that shouldn't be needed since all the objects that would receive the signals are destroyed.

Should I use `deleteLater()` instead?


srcejon:
> Should I use `deleteLater()` instead?

I think delete should be fine, thanks.

PR #2123, opened 2024-05-24
Fix ggmorse paths on mac

srcejon:
Fix ggmorse paths for Mac build.


f4exb:
Not related to GGMorse but we have this recurring issue on the Mac build:
```
error: could not find svn for checkout of lame
```
Perhaps svn has to be installed in the dependencies.

f4exb:
Great! Let's see how far we get now

f4exb:
Well... for what concerns ggmorse it works

Issue #2059, opened 2024-04-07
[audiofifo] Use after free in AudioFifo::write


dforsi:
Steps to reproduce:

1. compile with Address Sanitizer
2. run in a terminal the command `ASAN_OPTIONS="detect_odr_violation=1,strip_path_prefix=$(pwd)/" build/sdrangel`
3. make sure that sdrangel has started with an empty window, if not close all widget and restart
4. click an RX device and choose Audio
5. close sdrangel
6. the terminal shows output containing the following error message
`AddressSanitizer: heap-use-after-free sdrbase/audio/audiofifo.cpp:80 in AudioFifo::write(unsigned char const*, unsigned int)`

Here  is the complete output:
```
$ ASAN_OPTIONS="detect_odr_violation=1,strip_path_prefix=$(pwd)/" build/sdrangel
2024-04-07 15:17:02.912 (I) SDRangel 7.19.1-99-g99b06c605 Qt 5.15.10 64b x86_64 Debian GNU/Linux trixie/sid DSP Rx:24b Tx:16b PID 143153
2024-04-07 15:17:02.912 (I) MainSettings::MainSettings: settings file: format: 0 location: /home/daniele/.config/f4exb/SDRangel.conf
2024-04-07 15:17:05.697 (I) PluginManager::loadPluginsDir: loaded plugin libantennatools.so
2024-04-07 15:17:05.700 (I) PluginManager::loadPluginsDir: loaded plugin libaprs.so
2024-04-07 15:17:05.701 (I) PluginManager::loadPluginsDir: loaded plugin libbeamsteeringcwmod.so
2024-04-07 15:17:05.702 (I) PluginManager::loadPluginsDir: loaded plugin libchanalyzer.so
2024-04-07 15:17:05.703 (I) PluginManager::loadPluginsDir: loaded plugin libchannelpower.so
2024-04-07 15:17:05.706 (I) PluginManager::loadPluginsDir: loaded plugin libdemodadsb.so
2024-04-07 15:17:05.708 (I) PluginManager::loadPluginsDir: loaded plugin libdemodais.so
2024-04-07 15:17:05.709 (I) PluginManager::loadPluginsDir: loaded plugin libdemodam.so
2024-04-07 15:17:05.710 (I) PluginManager::loadPluginsDir: loaded plugin libdemodatv.so
2024-04-07 15:17:05.711 (I) PluginManager::loadPluginsDir: loaded plugin libdemodbfm.so
[...]

PR #2141, opened 2024-06-01
Fix heap-use-after-free

dforsi:
This PR fixes issue #2059 but breaks the device pulldown because when changing the current device it doesn't add the new one.

AddressSanitizer: heap-use-after-free /home/sdrangel/sdrbase/audio/audiofifo.cpp:80 in AudioFifo::write(unsigned char const*, unsigned int) AddressSanitizer: heap-use-after-free /home/sdrangel/sdrbase/audio/audiofifo.h:49 in AudioFifo::fill() const

I tested the audio input with my radio and a websdr and the audio output only with the loudspeakers.


dforsi:
Steps to reproduce the heap-use-after-free in AudioFifo::write:79:

1. compile Sdrangel with Address Sanitizer
2. start Sdrangel with an empty configuration
3. add an Audio Input source
4. close Sdrangel

Steps to reproduce the heap-use-after-free in AudioFifo::fill:49:
1. compile Sdrangel with Address Sanitizer
2. start Sdrangel with an empty configuration
3. add an Audio Output sink
4. close Sdrangel

Issue #1957, opened 2024-01-21
Please update "Satellite Tracker" GOES server file location


gvimlag:
TITLE:
  Please update "Satellite Tracker" GOES data file location

SYSTEM:
```
  Hardware-0: x86 64-bit
  OS-0:       Debian 12.4 (bookworm) 64-bit
  SDR:        RTL-SDR V3
  SDRangel: (compiled from source following "https://github.com/f4exb/sdrangel/wiki/Compile-from-source-in-Linux")
    Version 7.17.3-9-g9bfeddfea
    Build info: Qt 5.15.8 64 bits
    DSP Rx 24 bits    Tx 16 bits
```

BUG/ISSUE:
On startup, SDRangel gives the following error message:
```
  Satellite Tracker
  Failed to download: https://celestrak.org/NORAD/elements/goes.txt
  Error transferring https://celestrak.org/NORAD/elements/goes.txt - server replied: Not Found
```

When I enter "https://celestrak.org/NORAD/elements/goes.txt" into a browser, I get:
```
  Server Error
  404 - File Not Found.
```
So this server address is incorrect.

Please update the location of the "goes" satellite file in the SDRangel Satellite Tracker database.

Thank you.

srcejon:
This was changed a few days ago: https://github.com/f4exb/sdrangel/pull/1954

You can also edit the TLEs yourself in the settings dialog.

PR #2151, opened 2024-06-04
Add a button to reset the list of TLEs to defaults compiled in Sdrangel

dforsi:
This PR adds a convenience feature for users experiencing issue #1957 and to also get the new TLE source added in PR #1954


plugins/feature/satellitetracker/satellitetrackersettingsdialog.cpp
srcejon, line 1: #include "qmessagebox.h" should be #include <QMessageBox>

Use "" for sdrangel includes and <> for system/library includes.

dforsi:
I fixed the wrong include (added by Qt Creator...) and I added missing parts of the help page.

For the Replay tab I reworded the tooltips but I'm not sure that I understood the purpose of the checkbox "Send time to map" because I can change the replay date, but the Cesium map still shows the current time.

PR #2184, opened 2024-06-23
Fix/cleanup

dforsi:
This PR:
- removes some unused code:
   - `plugins/feature/map/mapfmlistdialog.ui` doesn't seem to be used anywhere, however it would be interesting to see it implemented
   - `sdrgui/gui/rollupwidget.{cpp,h}' don't seem to be used anywhere (the class that is actually used is names `RollupContents` and its files are very similar to these deleted)
   - some calls to `unlock` and some includes related to QMutexLocker
- fixes some typos in the swagger documentation and in a readme
-  applies some changes made automatically by Qt Creator that don't seem to change the appearance of the widgets
   - fix of a class name
   - removal of unused resources
   - changes to fonts (font name,  size and wieght)

I'm not able to regenerate the `html2` docs maybe because the [generate.sh](https://github.com/f4exb/sdrangel/commit/32bb49b631fa00a429020abb48a40201ec7b3c7e) script doesn't seem up to date because it is trying to edit .yaml files under sdrbase but there are none.


f4exb:
@srcejon I am not sure plugins/feature/map/mapfmlistdialog.ui should be removed. Maybe you had in the plans to implement this and we should save the effort of making up the dialog (although git would still remember it)

f4exb:
Building files from the .yaml Swagger definitions is done following these instructions: https://github.com/f4exb/sdrangel/tree/master/swagger/docker

srcejon:
> @srcejon I am not sure plugins/feature/map/mapfmlistdialog.ui should be removed. Maybe you had in the plans to implement this and we should save the effort of making up the dialog (although git would still remember it)

I was planning to import data from www.fmlist.org - but the owner of that site seems to have gone quiet, so probably wont happen. As you say, I can always dig up from git later.

PR #2187, opened 2024-06-27
Fix/map feature

dforsi:
This PR:

- adds the transmitters.csv file to the resources and converts its encoding from ISO-8859-1 to UTF-8 to show the accented letters
- modifies the table of map items settings
  - to right align all numeric cells
  - to use a QSpinBox for the "Filter Distance" cells
  - to do a case insensitive sorting of column containing the labels

For the  QSpinBox of "Filter Distance" I set a minimum value of zero and a maximum of 10000 km; I used `setSpecialValueText()` to replace the zero with one space so the the look of the table isn't too much different from before (we can't use an empty string or a nullptr because they disable the replacement).

The case sensitive sorting is needed otherwise "Airport" and "Airspace" were coming after APTdemod.

### DIFF
--- a/plugins/feature/map/data/transmitters.csv
+++ b/plugins/feature/map/data/transmitters.csv
@@ -1834,7 +1834,7 @@ FM,Heart (Morecambe Bay) Windermere WINDERMERE,Heart (Morecambe Bay),102300000,5
 FM,Heart Hertfordshire Hertford Cole Green,Heart Hertfordshire,106900000,51.787539,-0.135924,71,0.25,,
 FM,Heart Hertfordshire Hitchin and Letchworth Letchworth - Briar Patch Lane,Heart Hertfordshire,106900000,51.967745,-0.241566,97,0.05,,
 FM,Heart Hertfordshire Stevenage Old Knebworth,Heart Hertfordshire,106700000,51.872155,-0.211306,121,0.1,,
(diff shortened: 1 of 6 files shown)


srcejon:
> For the QSpinBox of "Filter Distance" I set a minimum value of zero and a maximum of 10000 km; 

Half the circumference of the Earth is larger than that.

srcejon:
> I used `setSpecialValueText()` to replace the zero with one space so the the look of the table isn't too much different from before (we can't use an empty string or a nullptr because they disable the replacement).

How does a user enter "No filter"? Can they delete the characters or do they have to insert a space? If the latter, then it would be better to create a custom spin box, like DMSSpinBox etc.

dforsi:
> > For the QSpinBox of "Filter Distance" I set a minimum value of zero and a maximum of 10000 km;
> 
> Half the circumference of the Earth is larger than that, so probably should be a bit bigger.

I was just thinking of a big number, but we can set it to 40075+1

dforsi:
> How does a user enter "No filter"? Can they delete the characters or do they have to insert a space? If the latter, then it would be better to create a custom spin box, like DMSSpinBox etc.

Now, with the commit that adds `QAbstractSpinBox::CorrectToNearestValue` the user can delete the characters, or type a zero which disappears when the cell looses the focus.

f4exb:
@srcejon Hello Jon, can it be merged?

srcejon:
> @srcejon Hello Jon, can it be merged?

Yep

PR #2243, opened 2024-09-03
Fix warnings and android issues

DreamNik:
Here's fixes for building for Android platform:
1) Fix compilation errors and warnings.
2) Changes related to building for Android using clang, Qt6 and fresh 3rdparty libraries, but could be useful when building in other environments too.
3) Fix plugin loading on android.


f4exb:
Sorry this breaks Metis MISO I am going to revert this PR

plugins/samplemimo/metismiso/metismisoudphandler.cpp
```diff
@@ -622,8 +622,6 @@ void MetisMISOUDPHandler::processIQBuffer(unsigned char* buffer)
                 }
             }
 
-            sampleMic  = (int)((signed char) buffer[b++]) << 8;
-            sampleMic += (int)((unsigned char)buffer[b++]);
```
f4exb, line 626: This is a breaking change and I don't see the point of it.

DreamNik: The variable "sampleMic" isn't used anywhere, so compiler complained about it, and thus I've removed it.

srcejon: b is modified and used though.

DreamNik: Yes, you're right! Didn't noticed that.

PR #2250, opened 2024-09-10
Use new find_package(CUDA Toolkit)

GlassOnTin:
On current Ubuntu Noble, unable to compile against CUDA using default sdrbase/CMakeLists.txt due to unsupported find_package(CUDA 9.0).

Instead use:
https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html

See suggestions at:
https://stackoverflow.com/questions/68475502/c-with-cuda-project-in-cmake-gives-error-nvlink-fatal-could-not-open-input


srcejon:
Looks like that requires CMake 3.17 - Currently we have: cmake_minimum_required(VERSION 3.13.0) - So I guess that needs updating too.

PR #2254, opened 2024-09-17
Allow plugins to be compiled as static libs, part 2.

srcejon:
These are the plugins that have embedded .qrc data. The files needed to be renamed so they are unique.

ADS-B: Allow plugin to be compiled as static lib. Allow building without texttospeech, location or process.
VOR localizer: Allow plugin to be compiled as static lib.
SID feature: Allow plugin to be compiled as static lib.
Satellite Tracker: Allow building without texttospeech or timezone.
Map: Allow plugin to be compiled as static lib. Allow building without SSL.
APT: Allow plugin to be compiled as static lib.
Radio Astronomy: Allow plugin to be compiled as static lib. Allow building without process.
ChannelWebAPIUtils: Fix memory leaks.
DeviceSampleSource: Add signals that indicate when the position or direction of a device (or its antenna) has changed. This is to support per device position, which is useful for remote devices not at My Position (E.g. Kiwi's / SpyServers etc). More to come in a later patch.


srcejon:
Can we modify the SonarCloud rules? 

E.g. for:

[Use the init-statement to declare "deviceReport" inside the if statement.](https://sonarcloud.io/project/issues?pullRequest=2254&open=AZH_PO32Um1X5WafiY_D&id=f4exb_sdrangel)

we aren't using c++17 and the suggested fix is certainly not more readable, as claimed, IMO.

f4exb:
I was thinking of upgrading to c++17 as most compilers now are at least at this level

PR #2270, opened 2024-10-08
Remote TCP updates

srcejon:
Add support for public list of SDRangel Remote TCP servers that can be displayed on Map.

image

Add FLAC and zlib IQ compression.
Add IQ squelch for compression.
Add remote device/antenna position and direction reporting.
Add text messaging between client and server.
Add client, time and sample rate limits.
Add wss (WebSocket Secure) protocol support for WebAssembly clients.
Fix list of Kiwi servers in map feature.

Note this adds new package dependency on libflac-dev.


srcejon:
Added a PR for sdrangel-docker to add the new package: https://github.com/f4exb/sdrangel-docker/pull/56

Issue #2159, opened 2024-06-10
No blocking on the main thread


srcejon:
For the WASM port, we need to ensure that there is no blocking on the main thread, otherwise SDRangel will hang.

This is described more in the Qt docs here https://doc.qt.io/qt-6/wasm.html under "Multithreading" and here in the Emscripten docs https://emscripten.org/docs/porting/pthreads.html#blocking-on-the-main-browser-thread

So far, I see two main areas that need to change in SDRangel:

- DSPDeviceSourceEngine::startAcquisition/stopAcquistion

When a device start button is pressed, DSPDeviceSourceEngine::startAcquisition is eventually called and is executed on the main/GUI thread. Its code blocks the thread until the device is started:

    DSPAcquisitionStart cmd;
    return m_syncMessenger.sendWait(cmd) == StRunning;

However, on WASM, USB devices will unable to start, because libusb calls get deferred to the main thread, so we get a deadlock.

As a workaround, it seems the blocking can be removed with:

    DSPAcquisitionStart *cmd = new DSPAcquisitionStart();
    m_inputMessageQueue.push(cmd);
    return true;

and the RTLSDR driver (and a few others I've quickly looked at) don't appear to require it to block. 

Removing this blocking would potentially have a small benefit for all platforms - as it would remove the small GUI freeze that can sometimes be seen when one device is already running and the start button is pressed on another device that takes a little while to initialize.

The above might not be the best way to implement this change - it's just the first thing I tried - and similar would also need to be implemented for TX and MIMO devices - and also for the stopAcquistion method.

- DeviceInput::applySettings

Currently DeviceInput::applySettings is also often called from the main/GUI thread, and the libusb calls result in a hang. I worked around this in RTLSDRInput::applySettings by moving all of the rtlsdr driver calls in to a separate thread. Similar would need to be done for other devices.

Again, this change is probably beneficial to all platforms, as discussed a while back in this issue: https://github.com/f4exb/sdrangel/issues/1522

srcejon:
DSPAcquisitionInit can hang as well, the second time time an RTL SDR is started. Similar workaround applies.

And forgot to mention, there needs to be coded added to DSPDeviceSourceEngine::handleInputMessages() to handle those messages.

(RTL SDR driver crash on stop has been worked around by using rtlsdr_read_sync instead of rtlsdr_read_async)

srcejon:
[...]

Issue #2273, opened 2024-10-09
7.22.1 issue -modssb


DB-Billy:
modssb not working w64

srcejon:
Can you be more specific please (including screenshots)? Very quick check here and it looks ok.

DB-Billy:
![Στιγμιότυπο οθόνης (93)](https://github.com/user-attachments/assets/a55da826-81b5-45fd-8450-dcdb8196444c)
this 2 bars not working at all remain left

srcejon:
Ok, I can reproduce that.

f4exb:
`SSBModGUI::on_spanLog2_valueChanged` is missing from `SSBModGUI::makeUIConnections` therefore the allowed span remains at 0 and other buttons cannot be moved.

DB-Billy:
thank u for all 73

PR #2274, opened 2024-10-10
MainWindow: Add FSMs to avoid blocking on GUI Thread

srcejon:
This PR adds some FSMs to MainWindow in order wait for events, while not blocking on the GUI Thread as discussed in #2159 

DSPDevice*Engine: Add signals to indicate when commands have been processed.
DSPDeviceSourceEngine: Fix small memory leak.
DSPEngine::removeDeviceEngineAt: Remove wait to avoid blocking thread. Return QThread to get QThread :finished signal.
DSPEngine::addDevice*Engine: Don't call deleteLater for device*Engine, as these objects are deleted manually in MainWindow, which will crash if deleteLater called first.

I don't quite understand some of the code in MainWindow::sampleSinkCreate and MainWindow::sampleMIMOCreate. In the original code, it does:

    const PluginInterface::SamplingDevice *samplingDevice = DeviceEnumerator::instance()->getMIMOSamplingDevice(deviceIndex);

but then:

    const PluginInterface::SamplingDevice *selectedDevice = DeviceEnumerator::instance()->getRxSamplingDevice(selectedDeviceIndex);

Why getRxSamplingDevice rather than getMIMOSamplingDevice for the second call? I've left that as is for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -704,6 +704,7 @@ if (BUILD_GUI)
                             QuickWidgets
                             Svg
                             SvgWidgets
+                            StateMachine
                            OPTIONAL_COMPONENTS
                             WebEngineQuick
                             WebEngineCore
--- a/sdrbase/dsp/dspdevicemimoengine.cpp
+++ b/sdrbase/dsp/dspdevicemimoengine.cpp
@@ -1030,6 +1030,7 @@ bool DSPDeviceMIMOEngine::handleMessage(const Message& message)
     else if (DSPAcquisitionStop::match(message))
 	{
 		setStateRx(gotoIdle(0));
+        emit acquisitionStopped();
         return true;
 	}
     else if (DSPGenerationInit::match(message))
@@ -1053,11 +1054,13 @@ bool DSPDeviceMIMOEngine::handleMessage(const Message& message)
 	else if (DSPGenerationStop::match(message))
 	{
 		setStateTx(gotoIdle(1));
+        emit generationStopped();
         return true;
 	}
 	else if (SetSampleMIMO::match(message)) {
         const auto& cmd = (const SetSampleMIMO&) message;
         handleSetMIMO(cmd.getSampleMIMO());
+        emit sampleSet();
 		return true;
 	}
 	else if (AddBasebandSampleSink::match(message))
@@ -1194,6 +1197,7 @@ bool DSPDeviceMIMOEngine::handleMessage(const Message& message)
         BasebandSampleSink* spectrumSink = msg.getSampleSink();
         spectrumSink->stop();
         m_spectrumSink = nullptr;
+        emit spectrumSinkRemoved();
         return true;
     }
     else if (SetSpectrumSinkInput::match(message))
--- a/sdrbase/dsp/dspdevicemimoengine.h
+++ b/sdrbase/dsp/dspdevicemimoengine.h
@@ -356,6 +356,11 @@ private slots:
 
 signals:
 	void stateChanged();
+
+    void acquisitionStopped();
+    void sampleSet();
+    void generationStopped();
+    void spectrumSinkRemoved();
 };
 
 #endif // SDRBASE_DSP_DSPDEVICEMIMOENGINE_H_
--- a/sdrbase/dsp/dspdevicesinkengine.cpp
+++ b/sdrbase/dsp/dspdevicesinkengine.cpp
@@ -458,12 +458,14 @@ bool DSPDeviceSinkEngine::handleMessage(const Message& message)
 	else if (DSPGenerationStop::match(message))
 	{
 		setState(gotoIdle());
+		emit generationStopped();
         return true;
 	}
 	else if (DSPSetSink::match(message))
     {
         const auto& cmd = (const DSPSetSink&) message;
 		handleSetSink(cmd.getSampleSink());
+		emit sampleSet();
         return true;
 	}
 	else if (DSPRemoveSpectrumSink::match(message))
@@ -476,6 +478,7 @@ bool DSPDeviceSinkEngine::handleMessage(const Message& message)
 		}
 
 		m_spectrumSink = nullptr;
+		emit spectrumSinkRemoved();
         return true;
 	}
 	else if (DSPAddBasebandSampleSource::match(message))
--- a/sdrbase/dsp/dspdevicesinkengine.h
+++ b/sdrbase/dsp/dspdevicesinkengine.h
@@ -121,6 +121,10 @@ private slots:
 
 signals:
 	void stateChanged();
+
+	void generationStopped();
+	void sampleSet();
+	void spectrumSinkRemoved();
 };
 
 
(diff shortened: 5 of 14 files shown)


srcejon:
> > Don't call deleteLater for device*Engine, as these objects are deleted manually in MainWindow ...
> 
> I am just wondering why the objects are deleted manually in MainWindow. Note: this is probably old code when I had not much experience with Qt and threads so it can be challenged possibly.

I think previously it wasn't a problem, as the device removal all happen in one go. Now the FSMs return control part way through the removal it gives the deleteLater a chance to be called and then occasionally it would crash on the delete in MainWindow. This seemed to depend on whether I just closed the device or was exiting the program. Obviously feel free to change it however you want.

f4exb:
It seems I cannot close the TestSink device using the cross (X) icon.

srcejon:
Ok, I'll take a look

Not that in DeviceGUi I changed the code to ignore the close event, as otherwise we would get a double delete (as MainWindow tries to delete it as well)

f4exb:
It is a bit strange when I tried to reproduce the issue on a new instance of SDRangel the TestSink closed as it should.

srcejon:
> Ok, I'll take a look
> 
> Note that in DeviceGUi I changed the code to ignore the close event, as otherwise we would get a double delete (as MainWindow tries to delete it as well)

This is because all the subclasses of DeviceGUI call setAttribute(Qt::WA_DeleteOnClose, true);

> It is a bit strange when I tried to reproduce the issue on a new instance of SDRangel the TestSink closed as it should.

I just tried on Windows a few times and it worked OK.

srcejon:
> It seems I cannot close the TestSink device using the cross (X) icon. This is what I see in the log:
> 
> ```
> 2024-10-12 08:44:58.766 (D) DeviceGUI::closeEvent
> 2024-10-12 08:44:58.766 (D) MainWindow::removeDeviceSet: index: 1
> 2024-10-12 08:44:58.766 (D) RemoveDeviceSetFSM::stopAcquisition
> 2024-10-12 08:44:58.766 (D) DSPDeviceSinkEngine::stopGeneration
> ```

My log shows:

```
2024-10-12 08:28:08.304 (D) DeviceGUI::closeEvent
2024-10-12 08:28:08.304 (D) MainWindow::removeDeviceSet: index: 1
2024-10-12 08:28:08.304 (D) RemoveDeviceSetFSM::stopAcquisition
2024-10-12 08:28:08.304 (D) DSPDeviceSinkEngine::stopGeneration
2024-10-12 08:28:08.304 (D) DSPDeviceSinkEngine::handleInputMessages: message: DSPGenerationStop
2024-10-12 08:28:08.304 (D) DSPDeviceSinkEngine::gotoIdle
2024-10-12 08:28:08.306 (D) RemoveDeviceSetFSM::removeSink
2024-10-12 08:28:08.306 (D) DSPDeviceSinkEngine::removeSpectrumSink:  SpectrumVis
2024-10-12 08:28:08.306 (D) DSPDeviceSinkEngine::handleInputMessages: message: DSPRemoveSpectrumSink
2024-10-12 08:28:08.306 (D) RemoveDeviceSetFSM::removeUI
2024-10-12 08:28:08.306 (D) DeviceGUI::~DeviceGUI
2024-10-12 08:28:08.306 (D) DeviceGUI::~DeviceGUI: end
2024-10-12 08:28:08.425 (D) RemoveDeviceSetFSM::stopEngine
```

So after stopGeneration, I'd expect to see DSPDeviceSinkEngine::handleInputMessages to handle DSPGenerationStop.

If that message isn't handled, it will not generate the generationStopped signal, and thus the RemoveDeviceSetFSM will not progress - as it waits for it, because going on to the next state.

So I guess the question is why might DSPDeviceSinkEngine not be handling messages?

f4exb:
The problem only happens when there is a modulator e.g the SSB modulator. Firstly when I stop the TestSink the SSB modultor stops but when I start again I cannot see the SSB modulator start debug message:
```
2024-10-12 09:50:16.761 (D) TestSinkOutput::handleMessage: MsgStartStop:  stop
2024-10-12 09:50:16.761 (D) DSPDeviceSinkEngine::stopGeneration
2024-10-12 09:50:16.761 (D) DSPDeviceSinkEngine::handleInputMessages: message: DSPGenerationStop
2024-10-12 09:50:16.761 (D) DSPDeviceSinkEngine::gotoIdle
2024-10-12 09:50:16.761 (D) TestSinkOutput::stop
2024-10-12 09:50:16.761 (D) TestSinkWorker::stopWork
2024-10-12 09:50:16.761 (D) DSPDeviceSinkEngine::gotoIdle: stopping  SSBMod
2024-10-12 09:50:16.761 (D) SSBMod::stop

2024-10-12 09:50:35.647 (D) TestSinkOutput::handleMessage: MsgStartStop:  start
2024-10-12 09:50:35.647 (D) DSPDeviceSinkEngine::initGeneration
2024-10-12 09:50:35.647 (D) DSPDeviceSinkEngine::startGeneration
```

Then if I stop again and try to close with (X) I get to the situation above:
```
2024-10-12 09:52:18.240 (D) TestSinkOutput::handleMessage: MsgStartStop:  stop
2024-10-12 09:52:18.240 (D) DSPDeviceSinkEngine::stopGeneration

2024-10-12 09:52:28.185 (D) DeviceGUI::closeEvent
2024-10-12 09:52:28.185 (D) MainWindow::removeDeviceSet: index: 0
2024-10-12 09:52:28.185 (D) RemoveDeviceSetFSM::stopAcquisition
2024-10-12 09:52:28.185 (D) DSPDeviceSinkEngine::stopGeneration
```

srcejon:
Ok, I can see the problem with the SSB and AM modulators, - but not the AIS or 802.15.4 modulators...

srcejon:
This problem seems unrelated to this patch, as I think it also occurs the release build of 7.22.1. 

Both SSB and AM Mod had threading changes in 7.22.1, so possibly something there.

srcejon:
It looks like m_thread->wait(); in SSBMod::stop never returns.

f4exb:
also I forgot to mention that I was still on your branch at commit `e3ec8759e [srcejon/freq_scanner] Remove unused variable` so if you added something more I missed it. I am going to try on the latest master...

srcejon:
For AISMod, after we call:

    m_thread->exit();

m_thread->isFinished() returns true and m_thread->wait() returns immediately.

However, for SSBMod, this isn't the case. m_thread->isFinished() returns false and m_thread->wait() hangs.

It may be something to do with being an audio sink, which the AISMod isn't.

If I remove:

    DSPEngine::instance()->getAudioDeviceManager()->addAudioSink(m_source.getFeedbackAudioFifo(), getInputMessageQueue());
 
From SSBModBaseband::SSBModBaseband then the thread exits.

srcejon:
The thread isn't exiting, because it is blocked in SSBModBaseband::processFifo m_channelizer->pull() - it doesn't seem to return from that call.

If I check the Audio Input button in the SSBMod, I don't see anything in the spectrum, with the 7.22.1 release, as I do with 7.22.0. (Maybe related to #2273?)

f4exb:
> Maybe related to https://github.com/f4exb/sdrangel/issues/2273

I don't think so... More probably due to this commit: 75d40c8b6 as you suggested before. It does not seem possible to handle the thread life cycle in the start / stop methods. It has to be tightly linked to SSBModBaseband.

PR #2276, opened 2024-10-11
Pager notifications

### LINKED ISSUE #2211: Converting special characters in pocsag logger
Hi!

I'm using the pocsag modulator to log messages to a csv file. Sometimes the alpha message contains characters like ' " ' which messes upp the structure of the csv file. Same thing with commas ','.

Is it possible to somehow convert those characters to some other before writing to the log file?

Thanks!

srcejon:
Add notifications to pager. User specifies a reg exp to match against address or message. If matched: message can be spoken, a command or script can be run, message can be highlighted or plotted on Map if it contains coordinates.
Add option to ignore duplicate messages.

Should also fix #2211


f4exb:
Note: I have raised the duplication rate in Sonar that does not make sense to me anyway...

srcejon:
Yeah, I couldn't see what the problem was with this.

For the other patch - I can reduce some of it if you want (i.e. the connects in sample*Create) - but I was just keeping the code as was.

doftheworld:
The notifications match on the "message" column, which for me displays the raw data received before it is decoded into the alpha column. Can this either be changed to match on the "alpha" column or that option added?

Issue #2294, opened 2024-10-28
Latest server version that works is 7.22.0


f4exb:
It appears that the latest version to work with the server flavour of SDRangel is 7.22.0. Something happened in 7.22.1 probably due to my changes for issue #2159 that caused an issue in the initialization sequence.

srcejon:
I would guess the server version needs a FSM to wait for the sync signals.

srcejon:
To recreate the problem, I'm running one of the example swagger scripts:

    ./rx_test.py -c -D RTLSDR

Part of the problem is in:

    void MainServer::changeSampleSource(int deviceSetIndex, int selectedDeviceIndex)

It does:

    deviceSet->m_deviceAPI->setSampleSource(source);
    deviceSet->m_deviceAPI->loadSamplingDeviceSettings(m_mainCore->m_settings.getWorkingPreset()); // load new API settings

Previously, setSampleSource would block, but this is no longer the case, so we have a race condition. 

We need to wait for the DSPDeviceSourceEngine::sampleSet signal before calling loadSamplingDeviceSettings.

We could do this with a FSM, as in MainWindow, but as MainServer.cpp is a bit simpler, we can get away with a single-shot lambda:

        // wait for sample source to be set, before loading settings
        auto connection = new QMetaObject::Connection();
        *connection = connect(deviceSet->m_deviceSourceEngine, &DSPDeviceSourceEngine::sampleSet, this, [=]() {
            deviceSet->m_deviceAPI->loadSamplingDeviceSettings(m_mainCore->m_settings.getWorkingPreset()); // load new API settings

            // Notify
            emit m_mainCore->deviceChanged(deviceSetIndex);
            
            QObject::disconnect(*connection);
            delete connection;        
        });       

Similar would be needed in changeSampleSink and changeSampleMIMO.

Qt6 has Qt::SingleShotConnection which would be nicer, as we don't need to create/delete the connection. but not in Qt5.  Might be better to code as an FSM for consistency with MainWindow however. But waiting for that signal fixes the problem for me.

f4exb:
Yes, this does fix the issue

f4exb:
Since it works I think this issue can be closed. Implementing FSM for the server is another subject. So far this solution is good enough.

PR #2304, opened 2024-11-01
Server: Use Sink/MIMO signals rather than Source.

srcejon:
When creating a sink we need to wait for the signal from the sink engine rather than source. Likewise for MIMO. For #2294


f4exb:
Sure, I just missed it in the copy + paste

PR #2305, opened 2024-11-02
github action to build ubuntu 22.04 and 24.04 release .debs

srcejon:
Not sure if you want this as well, but this adds a github action to build release .debs for both ubuntu 22.04 and 24.04. I guess some people might find the latter a useful addition. 

github has a 20.04 runner, but that only has cmake 3.16 and we currently require cmake 3.17.

Had to upload LimeSuite to latest release so it builds on 24.04. Also needed to tweak debian/control - but it works on both versions of ubuntu.

Currently only setup to build on release tag creation - but could possibly be used instead of appveyor for CI. It builds both versions in 36 mins and perhaps might be more reliable and save you some b/w.


f4exb:
> Not sure if you want this as well, but this adds a github action to build release .debs for both ubuntu 22.04 and 24.04.

Yes, that's quite interesting. Not sure we would like to keep 22.04 for long though.

> but could possibly be used instead of appveyor for CI.

That would be nice. Appveyor is a pain... but we might still keep it for Docker. This build usually works better than the .deb one.

PR #2311, opened 2024-11-05
Windows code signing: Use test certificate 

srcejon:
Windows code signing: Use test certificate until release certificate is available.


srcejon:
> It seems we are back to: `Input required and not supplied: api-token`

I think pull requests don’t get access to secrets – otherwise I could submit a PR that could expose it without that PR having been accepted.

Probably need to wait the next build after merge.

PR #2328, opened 2024-11-12
DemodAnalyzer: Ensure sample rate is set in worker.

### LINKED ISSUE #2309: Unable to play .wav file generated from Demod Analyzer
Recorded .wav file from NFMDemod and saved as a ...wav file.
Unfortunately, unable to play back the file.
Using macOS Ventura 13.7.1, SDRangel 7.22.2
Tried playback using iMusic and Elmedia player. 
Looking to record the audio, not the IQ information.

srcejon:
DemodAnalyzer: Ensure sample rate is set in worker. Should fix #2309.

Unrelated to the bug, have added support in Wav File code for meta-data chunks.


mjanicki:
Thanks guys.

PR #2344, opened 2024-12-02
Windows code signing update

### LINKED ISSUE #1877: Code signing on Windows
Download & install on Windows would be a great deal more user friendly, if the installer was signed (to bypass the maze of dialog boxes Windows throws up encouraging you not to run unsigned code).

Normally, this has a cost of several hundred $ a year for the certificates, but I came across [signpath.io](https://about.signpath.io/product/open-source) who appear to offer a free signing service & certificate for open-source software - and having tried a couple of apps signed with it, it appears to help (the installers run straight away).

It seems they need the build process to be verifiable - and currently support appveyor - with github actions in preview version.

Are you happy if I get in touch, to see if they will support us?

srcejon:
Remove unzip step, as it looks like it's not needed. And increase timeout to 1hour.

Hopefully that should now work and we can close #1877

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -103,7 +103,7 @@ jobs:
           github-artifact-id: '${{ steps.upload-unsigned-artifact.outputs.artifact-id }}'
           wait-for-completion: true
           output-artifact-directory: 'build\signed'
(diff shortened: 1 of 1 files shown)


f4exb:
It looks like it is working!

PR #2369, opened 2024-12-28
right-click-enabler: Allow activation of the right click action by keyboard

tyrylu:
This allows for example, selecting the audio output device for visually
impaired.

### DIFF
--- a/sdrgui/gui/crightclickenabler.cpp
+++ b/sdrgui/gui/crightclickenabler.cpp
@@ -81,6 +81,17 @@ bool CRightClickEnabler::eventFilter(QObject *obj, QEvent *event)
             }
         }
     }
+    else if (event->type() == QEvent::KeyPress)
+    {
+        auto keyEvent = (QKeyEvent*) event;
+
+        if ((keyEvent->key() == Qt::Key_Menu && keyEvent->modifiers() == 0) || (keyEvent->modifiers() & Qt::ShiftModifier && keyEvent->key() == Qt::Key_F10))
+        {
+            emit rightClick(QCursor::pos());
+            keyEvent->setAccepted(true);
+            return true;
+        }
+    }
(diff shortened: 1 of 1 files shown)


f4exb:
Can you please explain what it is supposed to do?

tyrylu:
yYou mean that the commit message is not descriptive enough (even with an example of a fixed use-case)?
It solves the bug where the right click action is only activatable using the mouse, and that's bad, because, well, not everyone uses the mouse.

srcejon:
> yYou mean that the commit message is not descriptive enough (even with an example of a fixed use-case)? It solves the bug where the right click action is only activatable using the mouse, and that's bad, because, well, not everyone uses the mouse.

Looks like it allows use of the menu key or shift-f10 to emulate right mouse click on the focused widget. Shift-f10 is the standard windows keyboard shortcut to display a context menu.

f4exb:
@tyrylu It just misses which keys are to be used at least in the comments

Thanks for the explanation @srcejon I tried in Linux but mousing over a right click enabled button and pressing Shift+F10 just does nothing. Is it a Windows only feature?

f4exb:
> Oddly, no KeyPress event gets to the rightclickhandler for shift-F10.

That's what I noticed also by putting there a debug message.

f4exb:
My mistake... it seem it works (on Linux) by tabbing to the concerned widget and not by mousing over.

srcejon:
The problem with shift-F10 on Windows, it's that Qt automatically converts it in to a QEvent::ContextMenu - which we are currently filtering, because that was giving us pop-up menus when tap-and-hold is used to emulate right-clicks. So probably need to revisit that.

tyrylu:
And the menu key is expected to do the same, right. By the way, why is the handling custom and does not use the QT context menu event?

Issue #2218, opened 2024-07-22
Autoplay on startup


johanios:
Hi! I have put sdr angel to automatically start on boot on my android device.

I Wonder if it is possible to automatically push the play button in sdr angel after the app has started?

i first want the sdr to start, and then automatically start the frequency scanner. Is that possible to automate somehow?

When saving a configuration, the settings is stored but the app is always started with the SDR stopped.

Any ideas?

thanks!

f4exb:
In the Android context I don't know... One possibility would be to have an ancillary script started in parallel that would pool the web API until it responds and then start the various plugins via API.

srcejon:
> In the Android context I don't know...

Haven't actually tried the API on Android...

But this request has come up a couple of times I think, so perhaps it might make sense to have dedicated functionality for it. i.e. some option under preferences.

johanios:
> In the Android context I don't know... One possibility would be to have an ancillary script started in parallel that would pool the web API until it responds and then start the various plugins via API.

That's a good idea, and I had considered solving it that way myself. However, for some reason, the running instance of SDR Angel instantly crashes when I request http://192.x.x.x/sdrangel. It works fine when requesting http://192.x.x.x/, but as soon as I add /sdrangel or any of its subpaths, the app immediately crashes on Android.

Do you have any idea why this is happening or what I should troubleshoot?

Thanks!

srcejon:
If you know how to use remote debugging on Android with adb you can use 'adb logcar' to look at the log to see if that tells you anything.

f4exb:
it is unhappy about the Swagger framework maybe...

PR #2381, opened 2025-01-20
Add --start command line option to start all devices and features. 

srcejon:
For GUI version of SDRangel, add --start command line option to start all devices and features automatically.

#2218


Axpelle:
Kindly note that the "Start/Stop all devices in workspace" button does not affect the state of the Frequency Scanner module.  That module has to be restarted separately after launching SDRA. Is that by design?

PR #2414, opened 2025-03-06
Fix HackRF support

### LINKED ISSUE #2283: SDRAngel 7.22.1 could not work with hackrf one
I have a hackrf one cloned by the third party. It works with SDRAngel 7.22.0 and another sdr software very well. After I upgraded SDRAngel to 7.22.1, it have received no signals but only noise. Is there a bug with hackrf one in 7.22.1 ？

srcejon:
Revert order of applySettings and thread starting in hackrfinput.cpp. As is, HackRF does not initialise correctly and signal is corrupted.

Add 250ms delay before loading device settings, to allow devices to initialise, as there is a 100ms DeviceGUI::sendSettings/updateHardware.

Should fix #2283


srcejon:
> The build for Windows failed, should that be fixed before merge

It did, but seems unrelated to this patch. Seems like a failure building limesuite.

PR #2438, opened 2025-04-21
Also accept +0179+0001 as an acceptable gs232 format for reporting po…

mattsoftware:
Fixes some of the GS232 implementation for the rotor controller plugin, specifically for the EA4TX ARS-USB which uses the GS232A protocol.


srcejon:
I'd suggest commenting out the debug of received data - this will fill up logs with it when tracking, losing other more useful info.

mattsoftware:
Commented out the debug statements as per your suggestion. Thank you.

Issue #140, opened 2018-02-27
Mismatch in 'struct' vs. 'class'


gvanem:
I'm on a mission to try to build SDRangel on Windows using MSVC. Realising that this is next to impossible now (variable length arrays and what not) , I've settled on the next-best thing; the wonderful `clang-cl` drop-in replacement. But even this returns several warning this these (with `-W3`):
```
plugin/pluginmanager.h(59,58):  warning: struct 'DeviceUISet' was previously declared as a class
      [-Wmismatched-tags]
    void createRxChannelInstance(int channelPluginIndex, struct DeviceUISet *deviceUISet, class DeviceSourceAPI ...
                                                         ^
plugin/plugininterface.h(19,7):  note: previous use is here
class DeviceUISet;
      ^
```
I think that due to this mismatch, the MSVC linker fails to link several .DLLs (mismatch in C++ mangled name). But it seems GCC is more liberal here; producing the same C++ mangling for both versions.

What is your intention? `class` or `struct`? I get this warning from approx. 20 places!

*Edit*. Here is an example of a missing symbol at link-time:
```
chanalyzerplugin.obj : error LNK2019: unresolved external symbol 
public: static class ChannelAnalyzerGUI * __cdecl 
ChannelAnalyzerGUI::create(class PluginAPI *, class DeviceUISet *,class BasebandSampleSink *)
```
In MSVC linker this has a different ABI than:
```
public: static class ChannelAnalyzerGUI * __cdecl 
ChannelAnalyzerGUI::create(class PluginAPI *, struct DeviceUISet *,class BasebandSampleSink *)
```
which gets produced in `chanalyzergui.obj`.

f4exb:
As it is forward declared as a class in many places let's make it a class in its own right then. I guess gcc makes no difference because in C++ there is not much difference but it never hurts to be consistent. Moving to clang-cl vs MSVC is probably a good move.

f4exb:
Fix is now on dev branch

gvanem:
Perhaps keep all those `class XX;` forwards in one common `.h`-file?

> Moving to clang-cl vs MSVC is probably a good move.

Great! The biggest hurdle with MSVC was in fact code like this:
```c
GLfloat q3[2*(end -start)];
```
Could it be done using `alloca()` or some STL-code instead?

gvanem:
So far, I've identified these warnings regarding `-Wmismatched-tags` in other places:  
```
[...]

Issue #2481, opened 2025-07-05
RE: 384000 Hz Sample Rate for AudioInput & AudioOutput


JoeyTheGamer5670:
So, I tried the latest 7.22.8, and it disappointed me. The SR on AudioInput is still capped at 96kHz, and the audio preference for Audio Input and Audio Output are still capped at 192kHz, meaning AudioOutput is still capped at 192kHz, not the 384kHz I was hoping for, for both AudioInput ad AudioOutput.

srcejon:
I'd guess Qt/Windows is reporting a lower maximum sample rate (because I added it to the SDRangel code).

What formats does Windows say are reported under System > Sound ? All sound devices > Your device > Format

Refers to #2456

JoeyTheGamer5670:
Could be QT, Control Panel says both the Hi-Fi Cable Output and Input are selected to "2 channel, 384000Hz", or perhaps I should enable Exclusive Mode in the properties?

srcejon:
Looks like it might be an issue in Qt https://bugreports.qt.io/browse/QTBUG-133538

We should be able to upgrade to 6.9.1.

srcejon:
Try this: https://github.com/f4exb/sdrangel/actions/runs/16088153799/artifacts/3469832564

Make sure it says Qt 6.9.1 in the bottom right corner of the window.

JoeyTheGamer5670:
Okay, 6.9.1, good, but uhh, it reports 192000Hz, not the 384000Hz I was looking for.

srcejon:
https://github.com/qt/qtmultimedia/blob/dev/src/multimedia/audio/qaudioformat_p.h

Seems to be missing 384000.

JoeyTheGamer5670:
so what will you do now?

srcejon:
Have filed a Qt bug: https://bugreports.qt.io/browse/QTBUG-138366

RachelMisaka:
Does the winexe version release still using the Qt 6.7.3

<img width="412" height="25" alt="Image" src="https://github.com/user-attachments/assets/a06f148a-d6fa-4f6b-842a-5644f78bb447" />

 or I need to update it manually?

srcejon:
> Does the winexe version release still using the Qt 6.7.3

Yes. 6.9 doesn't help anyway and has another problem.

PR #2484, opened 2025-07-07
Fix type forward declaration

srcejon:
Fix type forward declaration to match definition. For #140


srcejon:
Hmm, looks like the Windows build failure is due to https://bugreports.qt.io/browse/QTBUG-137763 - might need to roll back from Qt 6.9.1.

f4exb:
@srcejon Hello Jon, do you want me to merge this one and you deal with Qt6 later?

srcejon:
May as well go back to the old version now, as it apparently didn't help with #2481 anyway.

PR #2516, opened 2025-09-19
RigCtlServer: Add missing enabled setting key.

### LINKED ISSUE #2515: RigCtl Server vs Cat Relay
I am trying to use RigCtl Server and I'd like to know if could exist any pharameter needed, because I am able to connect to SDR++ using Hamlib, but I am not able to connect to SDRAngel.
Is there any log, where I could find what causes the connection fail?

Image

telnet localhost 4532
Trying ::1...
telnet: connect to address ::1: Connection refused
Trying 127.0.0.1...
telnet: connect to address 127.0.0.1: Connection refused
telnet: Unable to connect to remote host

Image

Regards, Alberto (PU1ABT).

srcejon:
RigCtlServer: Add missing enabled setting key. Fixes #2515

### DIFF
--- a/plugins/feature/rigctlserver/rigctlserversettings.cpp
+++ b/plugins/feature/rigctlserver/rigctlserversettings.cpp
@@ -139,6 +139,9 @@ bool RigCtlServerSettings::deserialize(const QByteArray& data)
 
 void RigCtlServerSettings::applySettings(const QStringList& settingsKeys, const RigCtlServerSettings& settings)
 {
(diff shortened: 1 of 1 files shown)


srcejon:
Looks like we need an update for LimeSuite to fix the Mac build cmake error.

albertovm:
Thank you for attention to this error. Is there any way to know when Mac version will be updated and fixed? Best regards... PU1ABT.

PR #2518, opened 2025-09-19
Updating to latest version of LimeSuite, Opus & Codec2

srcejon:
Try updating to latest version of LimeSuite to fix Mac build error, caused by cmake files that do not work with version 4 of cmake.


srcejon:
Looks like Opus and codec2 needs an update too.

srcejon:
Need a patch for cm256cc first as well - https://github.com/f4exb/cm256cc/pull/20

srcejon:
Add one for dsdcc as well: https://github.com/f4exb/dsdcc/pull/67

PR #2522, opened 2025-09-22
Further updates for cmake 4 support

srcejon:
Update bladeRF, libiio, soapysdr, airspyhf, hackrf, uhd and x265 to recent versions with cmake 4 support.
Use forks of mbelib, ggmorse and airspy with cmake 4 support. Can switch back when PRs are merged upstream.
Install ncurses for Mac build needed by bladeRF-power

Don't have bladeRF, Pluto or Airspy, so can't test if they still work.


srcejon:
Windows is now failing as they've switched the runner to Windows 2025, which doesn't appear to have NSIS installed according to https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md

f4exb:
Thanks a lot for fixing the builds 😀

PR #2536, opened 2025-10-15
stable version of plot

Amin4424:


Amin4424:
stable version added to sdrangel

PR #2620, opened 2026-01-22
NCO: Use fixed-point phase accumulator and interpolation

srcejon:
For a long time, I was puzzled why I couldn't completely correct a frequency offset in some of the demods, with the delta F setting. I've been using the NCO class and it seems this can have a frequency error of 15 Hz at 48kHz.

So, the following PR updates the NCO to use a Q12.20 fixed-point phase accumulator and linear interpolation, in order to improve it.

After writing this, I then noticed that some demods actually use NCOF, rather than NCO, which has a floating-point phase accumulator, which gives better frequency resolution than NCO.

Obvious question then, is which is better to use. Here's a comparison showing the spectrum for 5 different implementations, with a frequency of -15563Hz at 62,500Sa/s:

1) NCO from SDRangel
2) NCOF from SDRangel
3) The updated NCO in this PR using fixed-point accumulator and linear interpolation
4) All float implementation that calls sinf/cosf rather than using a LUT
5) All double implementation that calls sin/cos

image

And here are some performance figures for calling c *= nextIQ() in a tight loop:

NCO: 2.32 ns/itr
NCOF: 2.39 ns/itr
Fixed-point + interpolation: 2.41 ns/itr
Float: 5.16 ns/itr
Double: 9.62 ns/itr

Given that the fixed-point implementation has lower spurs and is around the same speed as NCFO (although we have the addition of interpolation, there's no need for the while loops in nextPhase), I think it may make sense to replace uses of that class with this as well, unless you see any other issues.

As a side-note: reducing table size to 11-bit increases spurs to -128dB. Increasing to 13-bit decreases to -152dB.

And as an implementation note, using std::fma for interpolation was slower with MSVC that just plain C code, sometimes much slower when without inlining and AVX2 support, as it results in a function call.

### DIFF
--- a/plugins/channelrx/demodinmarsat/inmarsatdemodgui.cpp
+++ b/plugins/channelrx/demodinmarsat/inmarsatdemodgui.cpp
@@ -506,8 +506,8 @@ QString MultipartMessage::decodeAddress(QString serviceCode, QString addressHex,
 
             if (coordinates)
             {
-                int lat1Deg = west ? -latDeg : latDeg;
-                int lon1Deg = south ? -lonDeg : lonDeg;
+                int lat1Deg = south ? -latDeg : latDeg;
+                int lon1Deg = west ? -lonDeg : lonDeg;
                 int lat2Deg = lat1Deg + latExtentEast;
                 int lon2Deg = lon1Deg + latExtentNorth;
                 *latitude = lat1Deg + lat2Deg / 2;
@@ -545,8 +545,8 @@ QString MultipartMessage::decodeAddress(QString serviceCode, QString addressHex,
 
             if (coordinates)
             {
-                int lat1Deg = west ? -latDeg : latDeg;
-                int lon1Deg = south ? -lonDeg : lonDeg;
+                int lat1Deg = south ? -latDeg : latDeg;
+                int lon1Deg = west ? -lonDeg : lonDeg;
                 QGeoCoordinate centre(lat1Deg, lon1Deg);
 
                 *latitude = lat1Deg;
--- a/plugins/channelrx/demodinmarsat/inmarsatdemodsink.cpp
+++ b/plugins/channelrx/demodinmarsat/inmarsatdemodsink.cpp
@@ -65,7 +65,7 @@ FrequencyOffsetEstimate::FrequencyOffsetEstimate() :
     FFTFactory *fftFactory = DSPEngine::instance()->getFFTFactory();
     m_fftSequence = fftFactory->getEngine(m_fftSize, false, &m_fft);
     m_fftCounter = 0;
-    m_fftWindow.create(FFTWindow::Rectangle, m_fftSize);
+    m_fftWindow.create(FFTWindow::Hanning, m_fftSize);
 }
 
 FrequencyOffsetEstimate::~FrequencyOffsetEstimate()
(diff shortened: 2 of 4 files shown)


f4exb:
Considering that the original NCO although having some error on the frequency but with -169.6 dB spurs that seem quite dithered I think we should keep it and have the fixed point phase with interpolation as another option not definitely deleting the original NCO.

Edit: I'd rather replace the NCOF that has much worse SFDR.

srcejon:
> Considering that the original NCO although having some error on the frequency but with -169.6 dB spurs that seem quite dithered I think we should keep it and have the fixed point phase with interpolation as another option not definitely deleting the original NCO.

The fractional NCO can reproduce the same output as the integer version, by just initialising the fractional part of the phase increment to zero, so I could add a parameter to support it. (Not sure if it would actually be useful in practice...)

f4exb:
I was just looking at the difference between plots 1 and 3 in the graphs above. The most significant visual difference is that in plot 1 the tiny spikes you see in 3 seem to be dithered and that probably explains the better SFDR in plot 1. I can't really explain why. Maybe adding some randomization possibly on the least significant bit of fracBits (i.e. m_phase) would achieve the same effect.

Edit: risking an explanation on why the present NCO achieves better SFDR... Although the phase is only on 12 bits the points on the sine curve given by the LUT are exact. When you do interpolation you are not exactly on the sine curve anymore. However randomizing the LSB of the 20 bit fractional part of the phase may statistically blur the error and eventually improve the SFDR.

srcejon:
> The most significant visual difference is that in plot 1 the tiny spikes you see in 3 seem to be dithered and that probably explains the better SFDR in plot 1.

The noise floor you see in plot 1 is I think a result of the windowing function used for the periodogram (hann).

The spurs in plot 3 only appear when there is a fractional part to the phase increment. If I choose the frequency so that it is the same as the NCO is effectively generating for plot 1, then you get identical output and no spurs.

I just tried adding some code to support dithering - this can be used to increase the noise floor to hide the spurs, but doesn't help reduce them, as I believe they're a result of the LUT rather than the phase accumulator.  

I think theoretically the spurs should be at 12dB per bit of the LUT, which is -144dBc (and that is exactly what we are getting as the peak power is 4dB). Doubling the LUT size reduces the spur to -152dB - so an option, but not sure if necessary.

I can push the code with optional dithering if you want it, but not sure if useful.

f4exb:
> The noise floor you see in plot 1 is I think a result of the windowing function used for the periodogram (hann).

Ah... so you think that this is an artifact of the visualization not from the signal so what we see is not exactly what we get correct?

srcejon:
I don't know enough about the intricacies of FFTs really, but yes. If I'd picked a frequency that was periodic within the FFT length the noise floor would be much lower.

f4exb:
It makes sense... For the dithering since you did the job maybe we should keep this option.
~Also not sure it is better but there is the option to do a weighted average for the interpolation~   i.e.
```c++
return m_table[i]*(1.0 - frac) + m_table[j]*frac;
```
This is exactly the same. Yours is more computationally efficient.

Edit2: if anything is to be improved going to Q14.18 (as done in AD DDS chips) makes a real change at the expense of a larger LUT. However one could use a quarter-wave LUT with a slightly more expensive LUT retrieval algorithm.

f4exb:
Considering the last conclusion, merging as the present implementation seems good enough. Later improvements may be done if the need arises

PR #2651, opened 2026-02-22
Voice Activity detection for Frequency Scanner in SSB modes

f4exb:


srcejon:
I think you need to add the spacers back in:

image

PR #2657, opened 2026-02-27
NAVTEX: Flush CSV after each message

rhbvkleef:
I'd like to watch for Navtex messages in an external application, though the file is only flushed whenever I restart SDRAngel or change the CSV writing settings. I'd like it to write out after each message so that I can continuously watch for new messages.

As such, I've added a flush after writing a message to the log file. This is not a huge burden on the filesystem as messages typically take around a minute to be received, and no less than 10 seconds.


rhbvkleef:
Off course ideally, I'd like to be able to extract the Navtex messages from the API, but that would probably involve more than a simple change.

PR #2706, opened 2026-04-20
NFMDemod: Correct FM scaling.

srcejon:
This PR should correct the FM scaling in the NFM Demod, so that the amplitude of a demodulated full scale signal is (-1,1) rather than (-0.5,0.5).

I was testing NFMMod to NFMDemod using the tone generator in the NFM Mod, whose amplitude is (-1,1), but only getting (-0.5,0.5) out of the demod, even though deviation was identical.

The problem I think is that m_fmDeviation is the full deviation rather than peak, so the factor of 2 isn't needed when calling m_phaseDiscri.setFMScaling.

We have a bit of an inconsistency between the different modems, where m_fmDeviation is sometimes the full deviation and sometimes peak.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -49,22 +49,21 @@
       "binaryDir": "${sourceDir}/build",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "RelWithDebInfo",
-        //"CMAKE_BUILD_TYPE": "Debug",
         "DEBUG_OUTPUT": "ON",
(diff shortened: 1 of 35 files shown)


srcejon:
The M17 demod had a factor of Pi in the call to setFMScaling, so I've removed that as well.

And I also updated the docs and GUI for some of the other modems that use FM to make it clear that the GUI setting is peak deviation, and so they are consistent with each other.

PR #2712, opened 2026-04-25
Use Qt6.2 compatible qt_add_plugin in CMake

tammojan:
I realize that Qt6 support is still experimental. This change may save someone else some time in the future. Debian 12 (Bookworm) ships with Qt 6.4.2. On that version, `qt_add_plugin` had no positional sources argument - it's simply ignored ([docs](https://doc.qt.io/archives/qt-6.4/qt-add-plugin.html)).  Funnily enough, SDRangel compiles and builds, but all plugins are compiled without sources, and thus don't load.

This change explicitly adds the sources to the plugin objects, which also works on newer Qt.

An alternative would be to require Qt >= 6.6, `qt_add_plugin` does have the sources argument ([docs](https://doc.qt.io/archives/qt-6.6/qt-add-plugin.html)).


srcejon:
You need Qt State Machines, which I don't think is available in Qt 6.2, is it?

tammojan:
Qt6StateMachine is installed by the debian package `qt6-scxml-dev`. So it is part of Qt 6.2, just well hidden.

srcejon:
> You need Qt State Machines, which I don't think is available in Qt 6.2, is it?

Ah, looks like it was added. I wouldn't recommend with such an early version of Qt 6 though. There have been many problems along the way.

tammojan:
Unfortunately I'm stuck with the Debian Bookworm computer for now. I can confirm that sdrangel compiles and runs with Qt6.2 after I apply this patch.

PR #2718, opened 2026-05-05
Make external libsigmf build on macos

tammojan:
The fork of libsigmf has diverged a lot from the upstream. The proper way to fix the shared library version of libsigmf on macos would be to merge the shared library fixes in the [new-namespaces branch](https://github.com/f4exb/libsigmf/tree/new-namespaces) with the upstream main. I've tried to do this merge but didn't succeed. The error in the underlying flatbuffers can just be ignored, that's what this patch does.
The build error this works around is:
```
In file included from /path/to/opt/libsigmf/external/flatbuffers/include/flatbuffers/idl.h:25:
/path/to/opt/libsigmf/external/flatbuffers/include/flatbuffers/flatbuffers.h:1720:25: error: definition of implicit copy constructor for 'TableKeyComparator<reflection::KeyValue>' is deprecated because it has a user-provided copy assignment operator [-Werror,-Wdeprecated-copy-with-user-provided-copy]
```
I know several people who would really appreciate sigmf support on MacOS, so it'd be great to have this in the official builds.


srcejon:
The mac build is failing because of cmake 4. This needs to be fixed for Ubuntu 26 as well, so we probably do need to merge some of the upstream changes in.

tammojan:
Thanks for merging! Before a release, we should check whether libsigmf.dylib is installed in the correct location; in a local build it still needed to be copied manually. Maybe this should be fixed by `MACOS_EXTERNAL_LIBS_FIXUP`, I don't understand this logic yet.

srcejon:
> Thanks for merging! Before a release, we should check whether libsigmf.dylib is installed in the correct location; in a local build it still needed to be copied manually. Maybe this should be fixed by `MACOS_EXTERNAL_LIBS_FIXUP`, I don't understand this logic yet.

MACOS_EXTERNAL_LIBS_FIXUP should point to where the dylib is built, so it can be found by the packaging script that collects all the required dylibs.

tammojan:
> MACOS_EXTERNAL_LIBS_FIXUP should point to where the dylib is built, so it can be found by the packaging script that collects all the required dylibs.

Thanks! In our local build we skipped the packaging script and just run `./sdrangel`, so it makes sense that we run into this problem. So probably the build is ok now :)

srcejon:
I think the sed command in this patch is what is causing the Debian builds to fail.

f4exb:
Trying to fix it on the fix-flac-build branch (concerns libsigmf also) because it makes the Debian builds fail also. However I think it has come to a point where some upgrades in my branch of libsigmf are necessary.

srcejon:
> Trying to fix it on the fix-flac-build branch (concerns libsigmf also) 

What error are you trying to fix there? To me it looked like copilot was misdiagnosing a FLAC warning, when it was sed that was failing (but could be wrong).

tammojan:
Not sure we're still discussing this pull request here since it's been merged for a while. I'm sorry if it broke a build; the sed command looks plausible to me but maybe it uses a mac specific thing. I agree that syncing the sigmf fork is a good thing to do, this sed-thing was a quick way to accomplish sigmf support on macos without going through the pain of merging a few years of divergence of the two libsigmfs.

srcejon:
> Not sure we're still discussing this pull request here since it's been merged for a while. 

The Debian and Mac builds that are broken are only built / tested once a release is made, not on every merged commit.

w1ebr:
I just tried SDRangel on MacOS 26 and got
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------
Process:             SDRangel [46929]
Path:                /Applications/SDRangel.app/Contents/MacOS/sdrangel
Identifier:          SDRangel
Version:             ???
Code Type:           ARM-64 (Native)
Role:                Default
Parent Process:      launchd [1]
Coalition:           com.f4exb.SDRangel [15602]
User ID:             501

Date/Time:           2026-06-07 17:03:09.9896 -0400
Launch Time:         2026-06-07 17:03:09.9552 -0400
Hardware Model:      Mac15,6
OS Version:          macOS 26.6 (25G5028f)
Release Type:        User

Crash Reporter Key:  1B045F16-91C8-B205-C500-1F9409109E30
Incident Identifier: 902080F9-42EC-4155-AE5D-4F0840BF3D78

Sleep/Wake UUID:       5CCC9A2A-358C-4DB9-8042-1E9CD397B72F

Time Awake Since Boot: 160000 seconds
Time Since Wake:       166 seconds

System Integrity Protection: enabled

Triggered by Thread: 0

Exception Type:    EXC_CRASH (SIGABRT)
Exception Codes:   0x0000000000000000, 0x0000000000000000

Termination Reason:  Namespace DYLD, Code 1, Library missing
Library not loaded: @rpath/liblibsigmf.dylib
---------------------------------------------------

When should I try again?
Thanks!

PR #2747, opened 2026-06-05
Fix libunwind detection on arm

df7cb:
The unwind code only properly worked on x64_64, but didn't get exercised on arm64.

Fix confirmed on https://buildd.debian.org/status/logs.php?pkg=sdrangel&arch=arm64


df7cb:
Fwiw the code is still somewhat wrong on architectures other than the two covered, it will run into linking errors on libunwind symbols.

The Debian package works around this by explicitly disabling libunwind elsewhere:
https://salsa.debian.org/debian-hamradio-team/sdrangel/-/blob/debian/master/debian/rules?ref_type=heads#L9-13

Issue #2761, opened 2026-06-06
Show Claimed Devices as Disabled Instead of Hiding Them?


rgetz:
Although it doesn't describe it in the [wiki](https://github.com/f4exb/sdrangel/blob/master/sdrgui/readme.md#12-create-new-receiver), currently, SDRangel excludes claimed/busy sampling devices from the device selection lists. As a result, when a device is in use, it disappears from the selection dialog entirely.

This can be confusing for users (at least it was for me) because it is difficult to distinguish between:
- A device that is disconnected or no longer detected.
- A device that is present but already in use by another device set.
- A device that was expected to be available but is not shown.

#### Proposed Behavior
Instead of filtering claimed devices from the enumeration lists completely, continue to display all non-removed devices and visually indicate that claimed/in use devices are unavailable (grey'ed out, and non-clickable, but still in the list).

Specifically:
- Continue to hide devices marked as removed.
- Include claimed devices in the RX, TX, and MIMO device lists.
- Display claimed devices in the device selection dialog as disabled/greyed out so they cannot be selected.
- Preserve the current behavior that prevents multiple claims of the same device.

The proposed behavior provides better visibility into device state and reduces user confusion (IMHO). Users can see that a device exists and is detected, but is currently unavailable because it is already claimed/in use, not look around trying to troubleshoot hardware.

Before sending the minor PR, I would like feedback on whether this is considered the desired UX for SDRangel, or whether the current behavior of hiding claimed devices is intentional and should be preserved.

PR is limited to (tested on linux x86 and ARM):
 sdrbase/device/deviceenumerator.cpp |    6 +++---
 sdrgui/gui/samplingdevicedialog.cpp |   31 +++++++++++++++++++++++++++++++
 2 files changed, 34 insertions(+), 3 deletions(-)

srcejon:
I've seen several people get confused by this in the past.

rgetz:
Thanks - If it's not just me :) - I will make a PR later this week.

rgetz:
Closing as this is now merged.

srcejon:
Reopening due to this issue: https://github.com/f4exb/sdrangel/pull/2770#issuecomment-5032265611

Some devices support multiple instances.

rgetz:
@srcejon - can you confirm that works for you? (it did for me on the file)

PR #2770, opened 2026-06-19
Show devices in use in device selection dialog

rgetz:
As discussed/described in #2761:

Previously, devices claimed by another device set were omitted from the device selection lists. Display these devices as disabled and marked "[in use]" instead of hiding them.

Tested on Linux (ARM and x86) on Qt6.

### DIFF
--- a/sdrbase/device/deviceenumerator.cpp
+++ b/sdrbase/device/deviceenumerator.cpp
@@ -332,7 +332,7 @@ void DeviceEnumerator::listRxDeviceNames(QList<QString>& list, std::vector<int>&
 {
     for (DevicesEnumeration::const_iterator it = m_rxEnumeration.begin(); it != m_rxEnumeration.end(); ++it)
     {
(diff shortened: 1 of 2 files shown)


rgetz:
Questions/review/feedback appreciated on UI:
- grey things out (done)
- add a tooltip (in use) - coded - but commented out / didn't include since it doesn't work on touchscreens, and when I did the next, it was duplication.
- add "[in use]" at the end of the name - to increase clarity/avoid questions of "why is my device grey'ed out" (done). I thought that was better user facing language (in use) than the internal "claimed") - but open for others opinions as well.

tested - when the dialog box is up, and moving up/down with arrows - with "disabled" Qt skips over it, so you can't select it. (which I think is OK/preferred - since you can never click "OK" on the dialog when a in use/claimed device is selected (which is existing functionality).

I'm not a fan of poking into the `QStandardItemModel` in the `QComboBox`, but it's the standard Qt5/Qt6 way of doing things (that I'm aware of).

srcejon:
There's a bug with this patch. It prevents you from opening multiple File Input / RemoteTCPInput / AudioInput / TestSources Devices, that all support multiple instances.

image

rgetz:
Sorry - I will have a look later tonight.

rgetz:
should be double fixed :)

PR #2774, opened 2026-06-24
Make usage of native flag optional while building

dawkagaming:
Hello,

I added a commit, which makes the usage of "native" march flag optional, as it can break the builds reproducibility
because it can change the output binary based on the micro-architecture of the local machine's CPU.

Thanks,
Dawid Kulas
SP9SKA

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,8 +33,7 @@ option(BUILD_GUI "Build GUI" ON)
 option(BUILD_BENCH "Build benchmark" ON)
 option(HIDE_CONSOLE "Hide console when running GUI on Windows" ON)
 option(BUNDLE "Enable distribution bundle" OFF)
-set(ARCH_OPT "native" CACHE STRING "Specify instruction set to use. Will be passed directly as `-march` or `/arch:` argument on supported compilers. \
-                                   'native' option will figure out host machine compatibilities and set flags accordingly (even with MSVC).")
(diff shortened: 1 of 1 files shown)


srcejon:
We already use -DARCH_OPT=nehalm to restrict the ISA for older CPUs

dawkagaming:
But it isn't about older amd64 CPUs, but also about other architectures.

srcejon:
What are you trying to set it to though?

Is it not the same as  -DARCH_OPT=""

dawkagaming:
"Will be passed directly as `-march`", so if it is set to "native" it will optimize the build for your CPU's march.
And no, setting it to "nephalem" won't work on arm64 for example.

We are building on more architectures than x86: <https://packages.debian.org/sid/sdrangel>

df7cb:
The context here is about *not* optimizing the build for a specific architecture as that violates distribution packaging policy. Unfortunately there doesn't seem to be a value that could be dropped into the existing `ARCH_OPT` flag to make it not use "native" and do a "normal" build instead. Hence we removed the whole line from CMakeLists.txt for Debian: https://salsa.debian.org/debian-hamradio-team/sdrangel/-/blob/debian/master/debian/patches/fix-rebuild.patch?ref_type=heads

This PR is about making that configurable so we can drop that patch.

df7cb:
Actually, `-DARCH_OPT=""` seems to work. :smile:

dawkagaming:
Okay, actually we can just use -DARCH_OPT="" to make it not set march option.

PR #2787, opened 2026-07-04
Improve PlutoSDR discovery robustness and identitection mgr

rgetz:
This change improves PlutoSDR device discovery reliability in multi-backend and multi-network environments by addressing two related issues in libiio-based scanning:

- Ambiguous device identity when the same hardware is exposed through multiple backends (IP, USB, etc.)
- Unstable or misleading URIs when mDNS (*.local) names are reused across different network segments

Together, these changes make device enumeration deterministic, uniquely identifiable, and more robust in heterogeneous network setups. (fixes issues where things like multiple plutos plugged in didn't work over IP, and if a local usb pluto would interfere with one using a USB<->Ethernet dongle).


gretel:
thanks, this is good for [tezuka](https://github.com/F5OEO/tezuka_fw), too!

PR #2814, opened 2026-07-19
plutosdr: Update RX gain limits dynamically from hardware

rgetz:
Query the PlutoSDR hardware for the current RX gain range and use it to configure the GUI gain control dynamically.

The AD936x gain limits vary with LO frequency. Previously the GUI used a fixed gain range, allowing users to select values that the hardware would reject after changing bands. The gain slider now refreshes its minimum, maximum, and step size whenever the device center frequency is updated.

Also change the gain setting serialization to use a signed integer so negative gain values are preserved for operating modes that support them.

Changes include:
- Add DevicePlutoSDRBox::getGainRange() to read in_voltage0_hardwaregain_available.
- Expose gain range through PlutoSDRInput.
- Refresh GUI gain limits when the device frequency changes.
- Avoid unnecessary widget updates when limits are unchanged.
- Store/restore gain as a signed value.
- Add default RX gain limit constants for fallback when the device cannot provide its range.


rgetz:
appveyor failing since `libairspyhf`'s `cmake/modules/FindUSB1.cmake` is incompatible with CMake 3.25+ because it calls `find_package_handle_standard_args()` with `FOUND_VAR LIBUSB_FOUND` while the package name is `USB1`. Modern CMake requires the found variable to match the package name (`USB1_FOUND`), causing configuration to fail.

PR #2816, opened 2026-07-20
Fix coverity int handling issues

rgetz:
This PR addresses several Coverity findings related to unsafe arithmetic, invalid state handling, and bit manipulation across multiple channel plugins.

The changes:
- Fix invalid bit extraction in ADS-B BDS 4,1 waypoint decoding. (another one that was not flagged last week).
- Prevent potential integer overflow during FT8 frequency conversion.
- Add validation around division operations in Noise Figure and Radio Astronomy calculations.
- Improve Radio Astronomy validation for calibration and sweep configuration edge cases.

These changes eliminate the reported Coverity issues while preserving existing behavior and making invalid runtime states fail safely.

Built all affected components, and run time tested the ADS-B ones.

### DIFF
--- a/plugins/channelrx/demodadsb/adsbdemodgui.cpp
+++ b/plugins/channelrx/demodadsb/adsbdemodgui.cpp
@@ -3970,7 +3970,7 @@ void ADSBDemodGUI::decodeCommB(const QByteArray data, const QDateTime dateTime, 
         c[2] = ((data[5] & 0x7) << 3) | ((data[6] >> 5) & 0x7);
         c[3] = ((data[6] & 0x1f) << 1) | ((data[7] >> 7) & 0x1);
         c[4] = ((data[7] >> 1) & 0x1f);
-        c[5] = ((data[7] & 0x1) >> 3) | ((data[8] >> 3) & 0x1f);
+        c[5] = ((data[7] & 0x1) << 5) | ((data[8] >> 3) & 0x1f);
         c[6] = ((data[8] & 0x7) << 3) | ((data[9] >> 5) & 0x7);
         c[7] = ((data[9] & 0x1f) << 1) | ((data[10] >> 7) & 0x1);
         c[8] = ((data[10] >> 1) & 0x3f);
--- a/plugins/channelrx/demodft8/ft8demodgui.cpp
+++ b/plugins/channelrx/demodft8/ft8demodgui.cpp
@@ -453,7 +453,7 @@ void FT8DemodGUI::on_applyBandPreset_clicked()
     int bandPresetIndex = ui->bandPreset->currentIndex();
     int channelShift = m_settings.getBandPresets(m_settings.m_decoderMode)[bandPresetIndex].m_channelOffset; // kHz
     int baseFrequency = m_settings.getBandPresets(m_settings.m_decoderMode)[bandPresetIndex].m_baseFrequency; // kHz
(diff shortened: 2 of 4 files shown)


srcejon:
>  run time tested the ADS-B ones.

Where did you get a BDS 4,1 message to test?

> Improve Radio Astronomy validation for calibration and sweep configuration edge cases.

The proposed fix prevents the spin box from scrolling through 0. It would be better if the parameters were validated in on_startStop_clicked().

rgetz:
> Where did you get a BDS 4,1 message to test?

I created a ADSB transmitter for a [GNU Radio presentation](https://www.youtube.com/watch?v=mbMvSSL3Yro) I did awhile ago. I modified that to create 4.1 messages. So I tested against synthetic messages. Sorry I should have mentioned that.
 
> The proposed fix prevents the spin box from scrolling through 0. It would be better if the parameters were validated in on_startStop_clicked().

My thinking was that a sweep step of 0 is not a valid configuration, so I prevented it from being entered rather than allowing an invalid state.

If the preferred approach is to validate in `on_startStop_clicked()`, should there also be user feedback (for example, a `QMessageBox()` explaining that the sweep step cannot be zero), or should the start simply be ignored? (to me - that's worse). I wasn't sure what the expected UX is for invalid parameter validation in this plugin.

srcejon:
> > Where did you get a BDS 4,1 message to test?
> 
> I created a ADSB transmitter for a [GNU Radio presentation](https://www.youtube.com/watch?v=mbMvSSL3Yro) I did awhile ago. I modified that to create 4.1 messages. So I tested against synthetic messages. Sorry I should have mentioned that.

Ah, ok. I don't think I've ever actually seen that message, so would have been interesting to see what aircraft was broadcasting it. Could do some more fun things with it if they did.

> > The proposed fix prevents the spin box from scrolling through 0. It would be better if the parameters were validated in on_startStop_clicked().
> 
> My thinking was that a sweep step of 0 is not a valid configuration, so I prevented it from being entered rather than allowing an invalid state.
> 
> If the preferred approach is to validate in `on_startStop_clicked()`, should there also be user feedback (for example, a `QMessageBox()` explaining that the sweep step cannot be zero),

Yep, I think that makes more sense.

rgetz:
> > If the preferred approach is to validate in `on_startStop_clicked()`, should there also be user feedback (for example, a `QMessageBox()` explaining that the sweep step cannot be zero),
> 
> Yep, I think that makes more sense.

Ok - will do later tonight. At the time - I didn't want to make major changes to UI.

I have two quick questions:

- I assume that if validation fails, after the QMessageBox is dismissed, the Start button should revert to the stopped state, since the sweep never actually starts?
- What message would you like displayed in the QMessageBox? I was thinking something along the lines of "Sweep step cannot be zero,", with only "OK", but I'm happy to follow whatever wording/style is preferred.

srcejon:
> Ok - will do later tonight. 

Thanks.

> I have two quick questions:
> 
> * I assume that if validation fails, after the QMessageBox is dismissed, the Start button should revert to the stopped state, since the sweep never actually starts?

Sounds sensible.

> * What message would you like displayed in the QMessageBox? I was thinking something along the lines of "Sweep step cannot be zero,", with only "OK", but I'm happy to follow whatever wording/style is preferred.

Yeah, sounds fine.

rgetz:
Updated.

PR #2820, opened 2026-07-21
Remove redundant ui null check in MeshcoreDemodGUI

rgetz:
The ui object is created during MeshcoreDemodGUI construction and remains valid for the lifetime of the widget. Remove the unnecessary null check to avoid misleading static analysis tools into assuming ui can be null.

This resolves a Coverity false positive where the later call to updateDechirpModeUI() was reported as a possible null dereference.


rgetz:
sorry - found another case of the exactly the same thing.

rgetz:
Maybe @gretel should also have a quick look.

PR #2821, opened 2026-07-21
APRS: Fix memory leak.

### LINKED ISSUE #2819: Confirm ownership model for `APRSGUI` / `APRSStation` memory management
While investigating Coverity CID 649241 (`RESOURCE_LEAK`), I found what appears to be a broader ownership issue and would appreciate confirmation before submitting a fix.

The Coverity report is for the following allocation in [`APRSGUI::handleMessage()`](https://github.com/f4exb/sdrangel/blob/6e2938a372466e02b48ef7b53a9295e99b531916/plugins/feature/aprs/aprsgui.cpp#L178):

```cpp
APRSPacket *aprs = new APRSPacket();
```

On the decode failure paths (`AX25Packet::decode()` or `APRSPacket::decode()` returning `false`), the packet is never deleted, so this appears to be a genuine leak.

While tracing ownership, I also noticed:

* `APRSPacket` objects are transferred to `APRSStation` via `APRSStation::addPacket()`, which simply appends the pointer to `m_packets`.
* `APRSStation` has no destructor and I could not find any code that deletes the stored `APRSPacket` objects.
* `APRSGUI` stores `APRSStation*` objects in `m_stations`, but its destructor only removes map items and deletes `ui`. I could not find any call to `qDeleteAll(m_stations)` or equivalent cleanup.
* `APRSGUI` is destroyed via `APRSGUI::destroy()`, so its destructor is expected to run.

I searched the tree for:

* `~APRSStation`
* `delete APRSPacket`
* `qDeleteAll(m_packets)`
* `qDeleteAll(m_stations)`
* `delete APRSStation`

and did not find any corresponding cleanup (but could be missing something).

Before submitting a patch, I'd like to confirm the intended ownership model:

1. Is `APRSGUI` intended to own the `APRSStation` objects stored in `m_stations`?
2. Is `APRSStation` intended to own the `APRSPacket` objects stored in `m_packets`?
3. If so, would adding an `APRSStation` destructor to delete `m_packets` and having `APRSGUI::~APRSGUI()` call `qDeleteAll(m_stations)` be the preferred solution?

If there is existing ownership or cleanup that I've overlooked, I'd appreciate being pointed to it.

srcejon:
Fix memory leak in APRS GUI. Fixes #2819

### DIFF
--- a/plugins/feature/aprs/aprsgui.cpp
+++ b/plugins/feature/aprs/aprsgui.cpp
@@ -175,9 +175,9 @@ bool APRSGUI::handleMessage(const Message& message)
     {
         MainCore::MsgPacket& report = (MainCore::MsgPacket&) message;
         AX25Packet ax25;
-        APRSPacket *aprs = new APRSPacket();
         if (ax25.decode(report.getPacket()))
         {
+            APRSPacket* aprs = new APRSPacket();
             if (aprs->decode(ax25))
             {
                 aprs->m_dateTime = report.getDateTime();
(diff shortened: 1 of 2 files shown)


rgetz:
LGTM

PR #2825, opened 2026-07-22
ft8: replace temporary malloc buffer with std::vector

rgetz:
Replace the temporary CRC working buffer in `LDPC::ft8_crc()` with a `std::vector<int>`.

The vector is value-initialized, preserving the existing zero-padding behavior, while `std::memcpy()` copies the input message into the front of the buffer. This removes the manual `malloc()`/`free()` pair, simplifies the implementation, and uses RAII for temporary storage.

This change also addresses a cppcheck warning about a possible null pointer dereference following malloc() failure. If memory allocation now fails, std::vector will throw std::bad_alloc rather than returning a null pointer that would be dereferenced. The algorithm and behavior are otherwise unchanged.

### DIFF
--- a/ft8/libldpc.cpp
+++ b/ft8/libldpc.cpp
@@ -35,7 +35,8 @@
 
 #include <stdio.h>
 #include <math.h>
-#include <stdlib.h>
+#include <cstring>
+#include <vector>
 #include "arrays.h"
 #include "libldpc.h"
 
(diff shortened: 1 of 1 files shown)


f4exb:
If memory allocation fails you are probably on the way to a whole system crash and reboot. This is why you may check the return... or not. But I do not mind the safety measure.

rgetz:
> If memory allocation fails you are probably on the way to a whole system crash and reboot. This is why you may check the return... or not. But I do not mind the safety measure.

Yeah - i will try to keep the value of the changes higher than the cost of review, if you think otherwise - let me know. There are still a few coverity issues I would like to fix, and then include weekly (or daily) submissions if that works for you. I will hold off on the cppcheck for now.

Issue #2830, opened 2026-07-23
Decide how to handle `-Wvla` compiler warnings


rgetz:
`CompilerOptions.cmake` (for 6+ years) currently enables:
```
add_compile_options(-Wall -Wextra -Wvla -ffast-math -fno-finite-math-only -ftree-vectorize)
```

Which makes the current build produces a number of warnings related to variable length arrays (VLA), for example, with :
```
gcc (Debian 14.2.0-19) 14.2.
/home/rgetz/github/sdrangel/plugins/channelrx/demoddatv/leansdr/sdr.h:2106:25:
warning: variable length array ‘data’ is used [-Wvla]

  std::complex<T> data[fft.n];
```
The question is not just "fix the warnings", since there are a few tradeoffs involved.

VLAs are a standard feature in C (introduced in C99), although their status has changed over time (optional in later C standards). However, they have never been part of standard C++. GCC and Clang support them as a compiler extension, which is why this code builds today, but they are not portable C++ and other compilers may reject them.

From a code quality perspective, replacing VLAs with standard C++ constructs would have some advantages:
- keeps SDRangel within standard C++
- improves compiler portability
- avoids relying on compiler-specific extensions
- removes a class of static analysis warnings
- avoids potentially large runtime-sized stack allocations

However, a blanket conversion may not always be the right answer. Some of these usages are likely in performance-sensitive DSP code, where the original author may have intentionally chosen stack allocation to avoid heap allocation overhead. Replacing every VLA with std::vector without considering the code path could introduce unnecessary allocations or affect performance.

Possible approaches:
- Leave -Wvla enabled and accept existing warnings : Treat these as known warnings. Don't change anything, but new occurrences of new warnings may be harder to notice because of existing noise. (`-Werror` isn't possible).
- Remove VLA usage where practical:  Convert general application code to standard C++ alternatives (std::vector, reusable buffers, etc.), while reviewing performance-sensitive paths separately.
- Allow intentional VLA usage : Keep the current implementation and disable `-Wvla` if the project considers compiler-extension usage acceptable.
- Handle case-by-case:  Remove accidental/unnecessary VLA usage, but allow documented exceptions where the allocation behavior is intentional and performance-related.

Before making any changes, I wanted to ask preferred direction first.
I searched for previous discussions, and could not find any.

srcejon:
[...]

PR #2841, opened 2026-07-25
step2 : reduce vlas 

rgetz:
This PR continues the effort to remove dependencies on compiler-specific variable length array (VLA) extensions within the DATV code.

The affected code uses temporary arrays whose size is determined at runtime. While GCC accepts these as an extension, they are not part of standard C++ and generate -Wvla warnings.

Replacing these temporary arrays with std::vector preserves the existing algorithms while reducing reliance on non-standard language extensions and improving portability.

These are small, mechanical changes with no intended functional impact.

Trying to do things as small atomic changes, so it's easier to bisect later if there are any issues.

Part of #2830


rgetz:
fix a space/tab issue that I didn't notice until I looked on the web view...

PR #2854, opened 2026-07-26
serialutil: Remove dead code in Windows COM port enumeration

rgetz:
Remove an invalid out-of-bounds array access that attempted to handle ERROR_INSUFFICIENT_BUFFER. The statement did not resize the fixed-size buffer and had no effect other than invoking undefined behavior.

The function only uses QueryDosDeviceA() as an existence check for COM ports, so removing the dead error-handling code does not change its behavior.


rgetz:
If you want to fix this in a more substantial way - and nack this - that's fine, this was just the minimal that removed the random dereference/undefined behavior crash (because I don't have a windows machine to test on).

PR #2860, opened 2026-08-01
gui: Explicitly call QSlider::setRange in LogSlider constructor

rgetz:
Cppcheck identified that LogSlider::LogSlider() calls the derived setRange(double, double) overload instead of QSlider::setRange(int, int) due to C++ name hiding.

The constructor's call to:
```
    setRange(0, 1000);
```
was therefore invoking the logarithmic range implementation with min == 0.0, resulting in a call to `log10(0.0)`. While most callers immediately replace the initial state by calling `LogSlider::setRange()` with a valid positive range, the constructor still performs an invalid mathematical operation during initialization.

Explicitly qualify the call as `QSlider::setRange(0, 1000)` to initialize the underlying slider without invoking the logarithmic overload. This eliminates the invalid-domain call, avoids undefined behavior from propagating exceptional floating-point values into the slider state, and makes the constructor's intent explicit.

This change has no functional impact on normal operation, but removes a latent initialization bug and prevents future regressions caused by overload hiding.


rgetz:
found the same - so fix them both in the same PR.

PR #2862, opened 2026-08-01
Remove remaining variable length arrays from demoddatv/ldpctool/generic.h

rgetz:
This PR removes all remaining variable length arrays (VLAs) from the ldpctool codebase, and the entire sdrangel codebase (that are flagged by warnings on my builds)

VLAs are a compiler extension and are not part of standard C++. While supported by some compilers (gcc, clang), they can trigger warnings, reduce portability, and create runtime stack allocation risks for larger buffers.

The changes replace VLA temporary buffers in the LDPC algorithm implementations with `std::vector` storage:

For cases with multiple temporary buffers of the same type, a single contiguous `std::vector` scratch buffer is used to preserve the original memory layout and avoid unnecessary allocations.

For `LambdaMinAlgorithm`, the conversion also updates `std::nth_element()` usage to operate on vector iterators instead of raw array pointers.

No algorithmic behavior is changed/intended; this is a portability and code quality cleanup to remove the final remaining VLA usage from the codebase.

### DIFF
--- a/plugins/channelrx/demoddatv/ldpctool/generic.h
+++ b/plugins/channelrx/demoddatv/ldpctool/generic.h
@@ -199,13 +199,17 @@ struct MinSumAlgorithm<int8_t, UPDATE>
 	}
 	static void finalp(int8_t *links, int cnt)
 	{
-		int8_t mags[cnt], mins[cnt];
+		std::vector<int8_t> scratch(3 * cnt);
+		int8_t *mags = scratch.data();
+		int8_t *mins = mags + cnt;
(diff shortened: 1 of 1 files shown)


f4exb:
Fails on Windows (MSVC). See: https://github.com/copilot/share/024a43ac-03c4-80d3-9902-be49e0504805

rgetz:
Sorry = tested on linux - will go back to the other way I was doing things, with multiple allocations. Testing now, and will push to be tested on CI later today.

rgetz:
Windows builds now.

PR #2871, opened 2026-08-03
fix various uninit member vars reported by Coverity and cppcheck

rgetz:
This PR addresses multiple uninitialized variable and member issues reported by Coverity and cppcheck.

The fixes ensure variables have defined initial values before use and prevent undefined behavior paths where optional inputs, initialization ordering, or unexpected execution paths could leave data uninitialized.

These changes do not alter normal execution paths; they only ensure deterministic behavior when values are accessed before explicit assignment.


rgetz:
accidentally fixed the same file twice, since the PR wasn't merged yet - my bad. dropped it here.

PR #2901, opened 2026-08-30
packet demod PACKETDEMOD_CHASE_MAX missing

tomasjura:
A declaration of a class static field in .h requires to have a definition in .cpp file. Without the patch the load of demod packet plugin fails with the error: 
`build_qt6/lib/plugins/libdemodpacket.so: sdrangel/build_qt6/lib/plugins/libdemodpacket.so: undefined symbol: _ZN17PacketDemodFramer21PACKETDEMOD_CHASE_MAXE`
The error is not visible during a compilation. It is visible only during the runtime.

### DIFF
--- a/plugins/channelrx/demodpacket/packetdemodframer.h
+++ b/plugins/channelrx/demodpacket/packetdemodframer.h
@@ -50,14 +50,15 @@
 // bit stuffing and flags is under 5000 symbols.
 #define PACKETDEMOD_SYM_HIST 8192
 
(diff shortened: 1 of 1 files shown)


tomasjura:
appveyor says: "Build execution time has reached the maximum allowed time for your plan (60 minutes)." Considering the proposed change complexity (2 line change in .h file), it needs another build attempt in appveyor. But some amendment in the appveyor is need.
The [2902](https://github.com/f4exb/sdrangel/pull/2902) suffers with the same problem.

f4exb:
Appveyor is crap ...

PR #2909, opened 2026-09-05
Fix uninitialized members in plugins/channelrx

rgetz:
yes - I know this is too large for a practical review. (my goal isn't to merge all these right now, unless you want to).

This is more of a discussion of how do you want these? (as one large thing, or multiple PRs, and if the latter - what size). I was using this specifically to test some integrations to OSH (Open Scan Hub). It's a little easier to do one thing for me - as I can resolve all the cppcheck issues, and then go see which are missing from coverity, and the squash them into the same file/commit, as to not have so many "fix" commits...

For OSH integration - this resolves/fixes/eliminates these issues from compiler warnings and cppcheck (544 issues)
https://openscanhub.fedoraproject.org/task/138770/log/fixed.html
(and I believe will reduce coverity issues). 

and adds these (none):
https://openscanhub.fedoraproject.org/task/137596/log/added.html

This (packit integration and eventually OSH) can be done on every PR, in a normal github workflow (so we do more per PR auditing, and don't let issues leak in as development happens). I'm just working with the upstream fedora/packit team to understand why the dual configuration of qt5 and qt6 isn't work. (right now, both use a qt6 configuration - ie has a bug).

If you do want to look at this as a large thing - it should be pretty straightforward - it's really just:
- setting default values in constructors (reordering a few to reduce warnings) 
  - removed `ifndef SERVER_MODE` in `DATVDemodSink::DATVDemodSink` since the var wasn't protected in the header, or code...
- added Q_FALLTHROUGH to quiet some compiler warnings
- removed unused variables
- added new way (qt6) of doing qt checkboxes, to eliminate deprecation warning in `chirpchatdemodgui.cpp` - mostly to test that the issue would disappear in OSH - I can drop that from the set of PRs if that's better - since that's the only thing that could change the way things work in a large way. (other than the small run time hit of clearing memory.


f4exb:
This is fine to do it in this PR

siteshwar:
@rgetz Thanks for the fixes and working with Packit team. I have added your name to the [list of contributors](https://github.com/openscanhub/openscanhub/wiki/Upstream-and-Fedora-Contributors) who have made fixes based on OpenScanHub reports.

PR #2911, opened 2026-09-06
Update rtl-sdr version

AsciiWolf:
Latest 2.0.3 release brings RTL-SDR Blog V4L support and other improvements.


srcejon:
Need to also update

    set(RTLSDR_TAG "v2.0.1")

In external/CMakeLists.txt, so it is updated for .debs and Windows / Mac.

At somepoint, we should change the flatpak and snaps to use external/CMakeLists.txt, so we only need to update in one place. I'll add an issue for that. #2914

PR #2921, opened 2026-09-13
mapsettingsdialog : Make API token configuration more obvious

rgetz:
Previously, users had to know where to register for each service and where to find or retrieve the required API key or token. Once configured, they then had to infer whether a key was working by waiting for a map to load, looking for a feature to become available, or noticing whether something started downloading in the background. This makes both initial configuration and troubleshooting more difficult than necessary.

Provide direct links to the relevant provider websites for registering, obtaining, and managing API keys and tokens. Add an explicit validation action that checks the configured credentials and immediately shows which keys or tokens are valid or invalid.

Implement asynchronous validation for the configured map and service provider credentials, with per-provider status displayed in the map settings dialog. Use provider-specific validation endpoints and response handling, including JSON validation where HTTP status alone is insufficient. Disable the validation button while requests are outstanding while keeping the dialog usable during validation.

enables this in the GUI (my keys/tokens removed, post validation):
sdrangel_map_settings

Figuring out where all these web sites are, which token to get, and if they are working - was at least painful for me.

This does add some ongoing maintenance to the map settings dialog. Provider APIs and authentication mechanisms can change, and a validation endpoint that works today may not continue to be suitable indefinitely.

The validation checks are therefore intentionally kept provider-specific and explicit, rather than trying to hide the differences behind a generic mechanism. If a provider changes its API or authentication behavior, the corresponding check will need to be updated along with the rest of the provider integration.

The benefit is that credential problems become immediately visible to users instead of being discovered indirectly through map or feature failures.


srcejon:
It's not just the Map with API keys and some are common between plugins, so I'd suggest moving the links & validation code in to a separate reusable class with separate methods for each key.

RE the GUI - can you just add click buttons with icons for each at the end of each row?

rgetz:
>It's not just the Map with API keys and some are common between plugins, so I'd suggest moving the links & validation code in to a separate reusable class with separate methods for each key.

Agreed. I was thinking of doing that - but I didn't want to restructure the existing things without a proof of concept that worked.

There are already provider-specific classes and functionality in SDRangel, but the current structure isn't particularly clean either. For example, Cesium-related functionality lives in plugins/feature/map/cesiuminterface.cpp, while CheckWXAPI lives in sdrbase/util/aviationweather.cpp, even though the credential/service functionality isn't necessarily specific to those locations. :)

Rather than creating another reusable class that follows the same pattern, I'll use this as an opportunity to look at the existing provider-specific functionality and move the pieces that belong at the reusable level into a common component.

That includes the registration/management links, credential handling, validation, and provider-specific URL/API details where appropriate. I'll also look at the existing functions in those areas and move them where it makes sense, rather than simply adding another layer on top of the current structure.

I think goal is to have one place for each provider's service/credential behavior, while keeping genuinely plugin-specific functionality where it belongs. That should also reduce the maintenance burden when provider APIs or authentication mechanisms change.

> RE the GUI - can you just add click buttons with icons for each at the end of each row?

Do you mean moving the validation status icon (✔/✗) to the end of each row? I actually had it there initially, but found that, at least for me, my eye didn't naturally go to it when the status appeared. By default the status is blank, and it only appears after clicking `Validate`, so I found it more noticeable at the start of the key/token.

Or do you mean replacing the single “Validate all Keys/Tokens” button with a validation button for each individual key/token, with an icon at the end of each row?

I'm happy to change it either way, I just want to make sure I understand the intended desire.

srcejon:
> Do you mean moving the validation status icon (✔/✗) to the end of each row?

The links "Thunderforest Website" "Register at Thunderforset" and "Get key". They can just be buttons with that text as tooltips.

rgetz:
>The links "Thunderforest Website" "Register at Thunderforset" and "Get key". They can just be buttons with that text as tooltips.

yeah - debated that - I chose text links rather than icon-only buttons because I wanted the purpose of each one to be immediately visible without requiring the user to hover or remember what an icon represents. (plus I hate picking out icons).

How about:
- https://lucide.dev/icons/external-link or https://lucide.dev/icons/globe
- https://lucide.dev/icons/user-plus
- https://lucide.dev/icons/key-round

My understanding is that lucide is license compatible with GPL3, unless there are other icons already used in the project?

and back to:
>can you just add click buttons with icons for each at the end of each row?

you mean just make the text box shorter, and 3 buttons at the end, rather than under?

srcejon:
Look OK to me, thanks. There are some existing icons in sdrgui/resources, but ok to add some new ones, if not already covered.

> you mean just make the text box shorter, and 3 buttons at the end, rather than under?

That would be my preference.

rgetz:
Makes sense - thanks for feedback - thoughts about having a 4th validate button for each row? (that saves you re-validating every token when you are struggling with one). (using something like https://lucide.dev/icons/refresh-ccw ). That was my only thing I really noticed when I was testing things. It also lends its self to the next/last question...

how programmatic do you want it? (provide row in the .ui, and call a function, and it gets added?) or it's all mostly in the `.ui` file? IMHO - for max reuse - I think it would be better to make it as programmatic as possible, and only have a cell in the `ui` to populate. (where the cpp creates the layout and inserts it) - that way it's always the same in every panel (scaling for the room it has).

and lastly - how secure do you want to display the tokens? is it better to display them as "****", "·····" or something? Some are free and throttled - some are free until you hit a limit, and they can cost you $/€ - I'm not sure I'm 100% comfortable displaying them as plain text when you just open the dialog - or maybe I have just been at too many CRA meetings at work recently. :)

rgetz:
after playing for a bit - a "common" way might be to put *all* the tokens / keys in a separate tab, and anytime someone wants to add this tab (for a key management aspect) - they can call it programatically (with the list of what services to display) and "add" it to their existing dialog (exactly what map display settings does), but unlike some others... ? (as opposed to line by line).

srcejon:
I have for some time been thinking we should have an API Keys dialog under Preferences menu, so a user could specify default values for all of their API keys (and only needs to do it once). Then, when the settings for a plugin are initialised to defaults, they use the default keys. The plugin's GUI would retain the existing widgets, so users can specify different per instance keys if needed. (I sometimes use different keys).

PR #2930, opened 2026-09-14
resolve uninitialized member variables in the plugins dir

rgetz:
This resolves another group of uninitialized member variables and few more minor issues in the `./plugins` directory

You  can see the some of the fixes https://openscanhub.fedoraproject.org/task/139978/log/fixed.html ; but there are alot of coverity issues that are not noticed by some of the open source tools.


rgetz:
If you want me to split things up - totally understand. Just let me know. This should resolve most of these sorts of issues (in the `./plugins` directory).

I go though things (a directory) with one tool, eliminating all the issues, and then iterate to the next tool, squashing things as necessary (since cppcheck, coverity, and compilers all complain about different things).

PR #2932, opened 2026-09-14
Position fallback options when GPS is unavailable

rgetz:
Provide users with alternative ways to obtain their position when an external GPS is not available.
* Use IP geolocation as a backup source for latitude and longitude when the current position is `0, 0`.
* Add a link to mapcoordinates.net so users can enter their address and obtain the latitude, longitude, and elevation values needed by SDRangel.

The IP geolocation result is only used as a fallback because it provides an approximate position. An existing user-entered or otherwise obtained position is never replaced.

The mapcoordinates.net link also makes users aware that the required position information can be obtained from a standard web site without needing a GPS or having to search for a suitable service themselves.

SDRangel_My_Position


srcejon:
> The mapcoordinates.net link also makes users aware that the required position information can be obtained from a standard web site without needing a GPS or having to search for a suitable service themselves.

That site is very ad-infested. 

Why not open the Map feature anyway? Hold shift and double click on the 3d map will give a marker with coords. (Yes, the 3D map is currently not displayed without an API key - a fix is coming for that soon).

rgetz:
I can update it to https://support.google.com/maps/answer/18539 if you think that's better? (or have any other preference) I just thought it was more direct, and you only do it once...

rgetz:
I think the issue - is I use a [pi-hole](https://pi-hole.net/) - and don't see many (any?) of the ads - if you think this is an OK idea - let me know which to use:
* https://www.gps-coordinates.net/
* https://developer.mapquest.com/documentation/tools/latitude-longitude-finder/

neither had ads for me.

srcejon:
> I think the issue - is I use a [pi-hole](https://pi-hole.net/) - and don't see many (any?) of the ads - if you think this is an OK idea - let me know which to use:
> 
> * https://www.gps-coordinates.net/

This is what it looks like without an ad blocker:

image


> * https://developer.mapquest.com/documentation/tools/latitude-longitude-finder/

This is better.

rgetz:
updated

Issue #2915, opened 2026-09-07
Closing Device/Channel/Feature via API while stack allocated dialog is open will crash


srcejon:
If a Device, Channel or Feature is closed via the API while any stack allocated dialogs are open, SDRangel will crash.

This includes all the calls to QMessageDialog::warning and things like QFileDialog dialog;

It can be fixed by allocating the dialogs on the heap instead. I'll add a MessageDialog class in sdrgui (Can't call it MessageBox on Windows as that collides with Windows API), that has MessageDialog::warning and similar methods, that allocate on the heap, so a simple replacement to fix those.

Adding this bug, as a few other things like QFileDialog, QColorDialog and QMessageBox::question will still need some work.

rgetz:
FWIW - I noticed while dialogs are open (either in the forefront or background), 
  - CTRL-Q does not work (I assume that it should do a cancel to everything that's open).
  - pressing on the same button to open the dialog, is a nop, rather than bringing said dialog to the forefront. (useful if it was behind windows).
  
I don't know if it's related or adjacent (my guess is not even close :) )- but if you are restructuring things - it might be worth a test.

Issue #2600, opened 2026-01-09
File duplication using FileSink API


volvico:
I use an Airspy HF+ as receiver. 
When doing : curl -X POST "http://127.0.0.1:8091/sdrangel/deviceset/0/channel/0/actions" -H "Content-Type: application/json" -d "{\"FileSinkActions\": {\"record\": 1},\"channelType\"

The FileSink begins to register in a first file, stops instantly (1ms) and restart in a second file. 

<img width="857" height="378" alt="Image" src="https://github.com/user-attachments/assets/b6898a69-c453-4a54-8251-5fdce04f7247" />

You can see that both files are created with 1 millisecond of difference. One is almost empty, while the other contains all the data. 

Thanks for your help.

PR #2934, opened 2026-09-15
Add MCP Server for Claude & ChatGPT

srcejon:
This PR adds an MCP Server feature to allow Claude & ChatGPT to control SDRangel and access audio, IQ and demod data.

Demo: https://youtu.be/G0FmiX7zwb8

While LLMs can use the existing Web API, MCP gives them additional information about how to use it to achieve different tasks (See bandguide.md which we should add to), which makes it more likely to work correctly and faster.

Although implemented as a feature, I made the GUI start it by default (hidden), so it should work out-of-the-box for complete beginners on Windows.

It's quite a large patch! The MCP Server uses the API to do most things and testing it showed up lots of bugs in the API: Missing/misspelt keys, missing reports, memory leaks, settingKeys issues etc. The LLMs benefit from the API keys having descriptions and ranges, so I've added all of those using AI, as they were mostly missing.

I've also added a few additional API endpoints for new functionality:

- spectrum/report - Gets spectrum measurements
- spectrum/data - Gets current spectrum data
- spectrum/history - Gets spectrum history (i.e. scroll buffer)
- spectrum/historyImage - Gets spectrum history as a PNG (Cheaper for LLMs to process than the numbers)
- spectrum/actions - For GUI operations such as autoscaling, clear and goto marker
- sdrangel/windows - Ability to arrange workspaces and hide windows (Note DeviceGUI, ChannelGUI, MainSpectrum and FeatureGUI now inherit from WorkspaceWindow, so some of the code for this is not duplicated). 

I've also added the ability to hide Device and Feature windows (so the MCP Server Feature started by default is hidden by default - feedback welcome on this approach and the AI button!). I've also added a new Windows menu, which allows hidden windows to be shown. 

Some other bug fixes included:

* QMessageBox crash when device/feature closed by API #2915 (Use new MessageDialog instead of QMessageBox)
* Duplicate IQ files #2600
* 3D spectrogram wobbling fixed
* Feature start/stop buttons not updating when started/stopped via API (Also added Feature::stateChanged signal to remove GUI polling)
* E4k RTL SDR frequency not being set on start, as PLL not settled.
* Test Source: Add period and pulse width settings
* In DeviceEngine, fix crash when a device is deleted, due to dangling pointer to a sink
* In Spectrum & Scope, fix crash on delete, due to dangling pointers
* Fixed deadlock in spectrum when settings applied via API


.github/workflows/sdrangel.yml
```diff
@@ -93,7 +93,7 @@ jobs:
           cmd "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
           choco install patch
           mkdir build && cd build
-          cmake .. -G "${{ matrix.config.generators }}" -DCMAKE_BUILD_TYPE=Release -DENABLE_QT6=ON -DARCH_OPT=SSE4_2 -DDEBUG_OUTPUT=ON -DENABLE_MIRISDR=OFF -DBUILD_SERVER=OFF -DCMAKE_PREFIX_PATH="C:\Qt\6.9.3\msvc2022_64;C:\Libraries\boost_1_73_0"
+          cmake .. -G "${{ matrix.config.generators }}" -DCMAKE_BUILD_TYPE=Release -DENABLE_QT6=ON -DARCH_OPT=SSE4_2 -DDEBUG_OUTPUT=ON -DENABLE_MIRISDR=OFF -DBUILD_SERVER=OFF -DBUILD_MCP_BUNDLE=ON -DCMAKE_PREFIX_PATH="C:\Qt\6.9.3\msvc2022_64;C:\Libraries\boost_1_73_0"
           cmake --build . --config Release --target package
```
rgetz, line 97: do you need to add the same to the other workflows (linux, linux-qt6.yml and coverity.yml?)

srcejon: Changed to building by default, as it should work on Windows, Linux and Mac.

rgetz: So - then this change is unnecessarily?

srcejon: I'm removed it in the latest patch

f4exb:
Since I've started working seriously with AI both for fun here and professionally and knew about MCP protocol I had the idea of adding the MCP capability to SDRangel. I haven't looked at your PR into detail but it would be great if I could ask the AI "Please build me a receiver with SDRangel that can do narrowband FM and also decode AIS messages with a map display" and it would perform it "magically" using the MCP facility (*). Anyway an excellent idea maybe worth a new major version (v8) as I think it is a breakthrough. In the first investigations I made I thought it would require Qt6 but I'm not sure it is strictly necessary. 

(*) note: in fact you probably need a dedicated agent to do this and this could be an ancillary project.

srcejon:
>  I haven't looked at your PR into detail but it would be great if I could ask the AI "Please build me a receiver with SDRangel that can do narrowband FM and also decode AIS messages with a map display" and it would perform it "magically" using the MCP facility (*).

This is what it does - take a look at the YouTube video: https://www.youtube.com/watch?v=G0FmiX7zwb8&lc

While it can work somethings out itself from the updated API descriptions & docs, there is a file [bandguide.md](https://github.com/f4exb/sdrangel/pull/2934/changes#diff-6c4a5025616e3f379e74a09e516bfb05080f6788517eada44444c70a24d9a6ff) where we can give it info about how to achieve specific task (E.g. what demods to use for what - what sample rates are required and so on). I've done that for some modems, but haven't yet done it for things like DATV etc.) - Not required, but makes it faster.

srcejon:
>  Anyway an excellent idea maybe worth a new major version (v8) as I think it is a breakthrough

I have another couple of big things coming in the couple of weeks :)

f4exb:
Just tell me when the PR is ready or just put it as work in progress

f4exb:
> take a look at the YouTube video: https://www.youtube.com/watch?v=G0FmiX7zwb8&lc

Awesome work! I am truly impressed :thumbsup:

srcejon:
> Just tell me when the PR is ready or just put it as work in progress

This MCP stuff should be ready, unless you want me to tweak how it's implemented.

The other things I just mentioned are completely different plugins.

f4exb:
> I have another couple of big things coming in the couple of weeks :)

OK, sorry for misunderstanding. So looking forward to other surprises then! :smile:

PR #2936, opened 2026-09-16
Fix Mac builds

srcejon:
Mac builds are broken on macos-14 - which now seems unsupported.

Update release to macos-15 and Qt 6.11.2 - and CI build to macos-26 with latest ffmpeg.


rgetz (approved):
Looks good to me - thanks for CI fix.

PR #2941, opened 2026-09-19
meshtastic: recover symbols with consistent +-1 FFT bin offset

jvn314:
Recover packets when symbols have a consistent ±1 FFT-bin offset.

On-air testing: decode success improved from 85% to 98%.

The explicit header uses SF-2 symbol bits, so each header value spans 4 FFT bins.

Peak FFT index  mod 4 = 0  -> peak is 1 bin low  -> +1 correction to full packet
Peak FFT index  mod 4 = 1  -> nominal -> no correction
Peak FFT index  mod 4 = 2  -> peak is 1 bin high -> -1 correction to payload only


jvn314:
Mac build should pass after PR #2936

jvn314:
closing because this patches the symptom and not the cause

